PR #4, opened 2019-12-04
GH-3 Fix listing edge routers for clusters. Fix api tests by copying …

plorenz:
…certs/etc from fabric


edge/controller/model/cluster_handlers.go
```diff
@@ -105,8 +105,8 @@ func (handler *ClusterHandler) HandleCollectEdgeRouters(id string, collector fun
 		if err != nil {
 			return err
 		}
-		association := handler.store.GetLinkCollection(persistence.FieldClusterEdgeRouters)
-		for _, edgeRouterId := range association.GetLinks(tx, id) {
+		edgeRouterIds := handler.store.GetRelatedEntitiesIdList(tx, id, persistence.FieldClusterEdgeRouters)
+		for _, edgeRouterId := range edgeRouterIds {
```
michaelquigley: This is the extent of the bug fix, yes? The rest of the PR is just making the ATs work again?

plorenz: Correct

rentallect:
Please merge `master` into this branch, then update the Jenkinsfile with that special tag on the "go test" line. I'd like to see your AT's run in CI.

edge/controller/model/cluster_handlers.go
michaelquigley: Seems pretty straightforward then, no?

plorenz:
> Please update the Jenkinsfile over in ziti-cmd with that special tag on the "go test" line. I'd like to see your AT's run in CI.

`go test` will only run tests from the local project, so we can't run the ATs from go-cmd. My plan is to add travis CI jobs for ziti-fabric and ziti-edge, similar to the one I added for ziti-foundation today, that will handle running tests.

PR #8, opened 2019-12-06
Check netlink errors not msgcount

scareything:
This PR fixes the netlink errors that cropped up recently in ziti-tunnel:

    failed to add local route: expected 1 netlink message, but got 0

This seemed to be benign at first glance, since the routes actually _were_ being added, but it was a problem for ziti-tunnel because it prevented the intercept setup from completing for a given service.

I didn't figure out why the problem started all of a sudden, but I'm guessing that it has something to do with vendored dependencies changing as we shifted to modules. Anyway, clients of the netlink package that we're using simply check the err returned by the Execute method.

### DIFF
--- a/tunnel/router/router_linux.go
+++ b/tunnel/router/router_linux.go
@@ -23,6 +23,7 @@ import (
 	"github.com/michaelquigley/pfxlog"
 	"golang.org/x/sys/unix"
 	"net"
+	"os"
 )
 
 // Add an address (or prefix) to the specified network interface.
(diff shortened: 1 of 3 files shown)


tunnel/router/router_linux.go
```diff
@@ -117,17 +117,16 @@ func nlAddrReq(localPrefix, peerPrefix *net.IPNet, ifName string, t netlink.Head
...
-	msgs, err := c.Execute(req)
-	if err == unix.EEXIST {
-		return nil
-	}
 
-	n := len(msgs)
-	if n != 1 {
-		return fmt.Errorf("expected 1 netlink message, but got %d", n)
+	_, err = c.Execute(req)
+	if err != nil {
```
plorenz: Do you need to check if the cast succeeded, or is Execute guaranteed to return errors of this type?

Issue #1, opened 2019-12-03
Create edge-router-policy entity type to replace clusters for determining edge router access


plorenz:
Currently clusters dictate which edge routers may be used by which services.

Edge router policies will instead link identities and edge routers. Identities and edge routers may have role tags defined on them, and edge router policies can be tied to identities and edge routers by both role tags and ids.

Edge routers can still be limited by services by specifying a edge router ids/role tags on the service directly.

If a service doesn't specify edge routers, all edge routers accessible to the identity dialing/binding the service will be available, otherwise it will be the intersection of the edge routers available to the service and the edge routers available to the identity

PR #14, opened 2019-12-11
 GH-1 Add support for edge router policies and service edge router li…

plorenz:
…mits


doc/breaking-changes-0.7.md
```diff
@@ -0,0 +1,99 @@
+ 
+
+This page discusses the changes that you need to be aware of when migrating your Ziti deployment from version 0.6.x to version 0.7.x
+
+# Theme 
+ * Ziti 0.7.0 replaces clusters with role attribute based policies.
```
andrewpmartinez: end period vs no end period

doc/breaking-changes-0.7.md
```diff
@@ -0,0 +1,99 @@
...
+  * identityRoles
+  * tags
+
+### Service Edge Router Roles
+Services now have a new edgeRouterRoles field. If set, this specifies which edge routers may be used for a service. This replaces the old cluster functionality. 
+
+### Edge Router Access 
+When a service is dialed or bound, which edge routers will be returned?
+
+  * If the service edgeRouterRoles are NOT set, then it will the set of edge routers that the dialing/binding identity has access to
```
andrewpmartinez: `it will the set` -> `it will be the set`

edge/controller/env/broker.go
```diff
@@ -353,6 +298,24 @@ func (b *Broker) sendToAllEdgeRouters(msg *channel2.Message) {
...
+		pfxlog.Logger().WithError(err).Errorf("could not get edge routers for session [%s]", sessionId)
+		return
+	}
+
+	for _, edgeRouter := range edgeRouterList.EdgeRouters {
+		edgeRouterEntry := b.edgeRouterMap.GetEntry(edgeRouter.Id)
+		if edgeRouterEntry != nil && !edgeRouterEntry.Channel.IsClosed() {
+			err := edgeRouterEntry.Channel.Send(msg)
+			if err != nil {
+				pfxlog.Logger().WithError(err).Errorf("could send session added message")
```
andrewpmartinez: "could send" to "could not send"

edge/controller/env/broker.go
```diff
@@ -513,58 +448,42 @@ func (b *Broker) sessionCreateEventHandler(args ...interface{}) {
...
-				}
-			} else {
-				pfxlog.Logger().WithField("cause", err).Error("error sending network session added, could not marshal message content")
-			}
-		}
+	if buf, err := proto.Marshal(sessionAdded); err == nil {
+		msg := channel2.NewMessage(SessionAddedType, buf)
+		b.sendToAllEdgeRoutersForSession(session.Id, msg)
+	} else {
+		pfxlog.Logger().WithField("cause", err).Error("error sending network session added, could not marshal message content")
```
andrewpmartinez: Logging messages in the format of `WithField("cause", err)` we should change them to `WithError(err)`. `WithError()` is simply `WithField("error", err)` but it more semantic and most of the other ziti codebase does it that way.

I simply didn't know about WithError till it was too late.

doc/breaking-changes-0.7.md
plorenz: Fixed

doc/breaking-changes-0.7.md
plorenz: Fixed

edge/controller/env/broker.go
plorenz: :+1: Fixed.

edge/controller/env/broker.go
plorenz: Fixed

edge/controller/model/session_handlers.go
```diff
@@ -151,16 +151,27 @@ func (handler *SessionHandler) HandleList(queryOptions *QueryOptions) (*SessionL
 	return result, nil
 }
 
+func (handler *SessionHandler) HandleListSessionForEdgeRouter(edgeRouterId string) (*SessionListResult, error) {
+	result := &SessionListResult{handler: handler}
+	query := fmt.Sprintf(`anyOf(apiSession.identity.edgeRouterPolicies.edgeRouters) = %v and `+
```
andrewpmartinez: mmm the query power!

edge/controller/env/broker.go
michaelquigley: I actually prefer `logrus.Errorf("message (%w)", err)`-style wrapped errors. I find both the messages, and the code easier to read.

edge/controller/model/session_handlers.go
plorenz: Glad you mentioned it, b/c I noticed the query was wrong in several ways. Had to add ability to check for empty sets in query lang, see netfoundry/ziti-foundation#7

edge/controller/env/broker.go
plorenz: I'll check it out. I'm dissatisfied with the current state of error handling in general. I've been using `Errorf("%+v", err)` in order to view stack traces when they are available. Will %w do that as well?

michaelquigley: Wrapping errors (so that they can be unwrapped later) is the primary reason for using `%w`. I believe the point of wrapping is so they can be unwrapped later. That may not help with stack traces. 

The `ziti-fabric` codebase still uses `%s` in a lot of places, but I plan on migrating them over to `%w` as I work through there.

Issue #18, opened 2019-12-18
De-stutter /ziti-edge/edge


andrewpmartinez:
There is no need for ziti-edge/edge/...

Remove the 'edge' folder to de-stutter.

PR #19, opened 2019-12-18
GH-18 removes ziti-edge/edge folder and moves all contents up 1 row

andrewpmartinez:
- Updates .gitignore files
- Fixes imports


andrewpmartinez:
> Why did all the permissions package imports get aliased to permissions2?

Goland mangled some imports. Added some names, changed some names. Not sure why.

I went through and fixed them all.

PR #24, opened 2019-12-31
Track shared routes

scareything:
This PR pays closer attention to intercept cleanup, and avoids the following error when `tunnel/router.RemoveLocalAddress` is invoked to remove a route that doesn't exist (or has already been removed):

    ERROR github.com/netfoundry/ziti-edge/tunnel/intercept/tproxy.(*tProxyInterceptor).Stop: failed to clean up intercept configuration for ssh-direct: failed to remove route for service ssh-direct: netlink receive: cannot assign requested address

The service poller has always tried to intelligently remove routes when services are removed from app-wans, but two scenarios thwarted its logic:

1. `ziti-tunnel` side-steps the route detection logic when it is shutting down. In this case, routes for services were unconditionally removed. This PR changes the service poller shutdown sequence so that route shared routes are always detected.
2. The tproxy interceptor kept track of UDP and TCP intercepts for a given service as two distinct services, which inadvertently led to `tproxy.StopIntercepting` removing the same route twice (once for UDP and once for TCP). This PR updates the tproxy interceptor so that it only attempts to remove each route once.


dovholuknf:
crud - I meant to leave a comment asking: Can you please add the details as to 'how' the cleanup works as some doc somewhere?

tunnel/intercept/svcpoll.go
```diff
@@ -21,58 +21,86 @@ import (
...
-			if _, ok := knownServices[edgeSvc.Id]; !ok {
-				addedServices[edgeSvc.Id] = edgeSvc
-				knownServices[edgeSvc.Id] = edgeSvc
+		select {
+		case <-time.After(pollRate):
+			continue
+		case s := <-sig:
+			log.Debugf("caught signal %v", s)
+			switch s {
+			case os.Interrupt, os.Kill, syscall.SIGTERM:
```
ekoby: any reason for variance in namespace? Also what about other terminating signals (QUIT, ABORT, SEGV)?

also can we handle suspend/wakeup (SIGSTOP/SIGCONT) sequence?

tunnel/intercept/interceptor.go
```diff
@@ -136,14 +136,15 @@ func (m LUT) GetByAddress(addr net.Addr) (*InterceptedService, error) {
 	return &s, nil
 }
 
-func (m LUT) GetByName(serviceName string) (*InterceptedService, error) {
+func (m LUT) GetByName(serviceName string) []InterceptedService {
+	r := make([]InterceptedService, 0)
```
plorenz: Best practice is to start with nil slice, since len and append both work with nil slices. 

`var r []InterceptedService`

At least that's what the linter always tells me :)

scareything: Done, thanks! golang is similar enough to C that I can sometimes forget where I am :)

tunnel/intercept/svcpoll.go
scareything: At one point we were only using the more portable `os` package, which only deals with signet and sigkill. Eventually sigterm was added, and I should've changed the others to `syscall` at that time. Done now.

As for stop/cont, we could do that but I'm not 100% clear on how useful it is?
- SIGSTOP would probably remove all intercepts but remain in the loop
- SIGCONT would probably `continue` the loop to ensure a quick poll for services after waking up (regardless of the poll rate)

ekoby: the process could be suspended and then terminated without giving you the chance to clear the routes. It is probably rare set of circumstances but it seems that it won't be hard to handle for completeness

PR #32, opened 2020-01-09
Remove Handle prefix from handler methods. Remove breaking changes do…

plorenz:
…cs, which are moving to ziti-cmd


michaelquigley:
Wouldn't it make sense to either use the `Handler` idiom throughout the codebase (both edge and fabric), or jettison it throughout the codebase?

plorenz:
> Wouldn't it make sense to either use the `Handler` idiom throughout the codebase (both edge and fabric), or jettison it throughout the codebase?

Yes, if they were following the same pattern, but I don't feel they are. If/when we put the mgmt channel in front of edge capability there will be a Handler layer in front of it, with appropriate HandleRequest methods. So in the same way that network doesn't have Handle prefixed everywhere, I don't think this code should either. 

That probably means they should not be called Handler either, but Controller is longer and Model is kind of overloaded. Can rename them when we come up with a good name :)

michaelquigley:
> That probably means they should not be called Handler either, but Controller is longer and Model is kind of overloaded. Can rename them when we come up with a good name :)

"Controller" is also overloaded.

Just want to keep making sure that we're thinking about the entirety of the architecture, not just the edge.

controller/model/associations_handlers.go
```diff
@@ -59,8 +59,10 @@ func (handler *AssociationsHandler) UpdateAssociations(store persistence.Store,
 			return linkCollection.AddLinks(tx, parentId, childIds...)
 		case AssociationsActionRemove:
 			return linkCollection.RemoveLinks(tx, parentId, childIds...)
-		default:
+		case AssociationsActionSet:
```
andrewpmartinez: Wasn't mentioned in the commit message. Was this on purpose?

plorenz: oh, yes. Small tidy. Make it explicit which actions are supported in case we ever add new ones

PR #33, opened 2020-01-09
Fix edge router roles check. services with role attributes were getti…

plorenz:
…ng empty edge router lists


controller/model/edge_router_handlers.go
```diff
@@ -158,7 +158,7 @@ func (handler *EdgeRouterHandler) ListForIdentityAndServiceWithTx(tx *bbolt.Tx,
 
 	query := fmt.Sprintf(`anyOf(edgeRouterPolicies.identities) = "%v"`, identityId)
```
plorenz: This is the actual fix, the rest is all adding a test

PR #35, opened 2020-01-10
Config store implementation. Fixes GH-17.

plorenz:


plorenz:
This is the basic implementation of a config store. It's not connected to services or identities yet. It allows configuring arbitrary JSON-like data, like tags excepted nested and assigning a name to it.

I will add ATs before the PR should be considered complete

PR #38, opened 2020-01-13
API Tests - Session Refactor

andrewpmartinez:
- API tests now have a session struct that is used to maintain tokens
- Moved authenticated requests to their own struct to scope requests
- Updated existing tests to use new pattern


tests/authenticate.go
```diff
@@ -0,0 +1,428 @@
...
+	if sessionToken, ok := bodyContainer.Path("data.token").Data().(string); ok {
+		sess.token = sessionToken
+	} else {
+		return errors.Errorf("failed to authenticate via UPDB as %v: failed to find session token", sess.authenticator)
+	}
+
+	if identityId, ok := bodyContainer.Path("data.identity.id").Data().(string); ok {
+		sess.identityId = identityId
+	} else {
+		return errors.Errorf("failed to authenticate via UPDB as %v: failed to find identity id", sess.authenticator)
```
dovholuknf: i tend to not like providing any kind of information as to why authentication fails. can we just say "could not log in"? or something?

dovholuknf: oh hey that's in a test. nevermind :)

plorenz:
I like it

Issue #30, opened 2020-01-09
Allow linking config to services by configurable type


plorenz:
* Add new configurables field to identity, to indicate which configurables the identity should recieve
* Add configurable -> configId mapping on service

When an identity retrieves services they should receive the configuration data for mapped configurables.

So given
  * config entry with id = "1" and data { dnsHostname: "ssh", port: 22 }
  * identity A has configurables = ["ziti-tunneler"] 
  * service ssh has configMappings { "ziti-tunneler" : 1 }

Then when identity A lists that services, it should have the following included in the S1 data
```json
    "config" : {
        "ziti-tunneler" : {
            "dnsHostname" : "ssh",
            "port" : 22
        }
    }
```

PR #45, opened 2020-01-17
Allow linking configurations to services. Fixes GH-30

plorenz:


controller/internal/routes/authenticate_router.go
```diff
@@ -102,3 +105,18 @@ func (ro *AuthRouter) authHandler(ae *env.AppEnv, rc *response.RequestContext) {
 
 	rc.RequestResponder.RespondWithOk(currentSession, nil)
 }
+
+func (ro *AuthRouter) mapConfigTypeNamesToIds(ae *env.AppEnv, values []string) []string {
+	var result []string
+	if stringz.Contains(values, "all") {
```
dovholuknf: let's say we internationalize this one day - would 'all' still be the special token? not sure if that's something to consider now or not but internationalization has been rolling around in my head for a while so i'm making the comment

controller/internal/routes/authenticate_router.go
```diff
@@ -102,3 +105,18 @@ func (ro *AuthRouter) authHandler(ae *env.AppEnv, rc *response.RequestContext) {
...
+		return []string{"all"}
+	}
+	for _, val := range values {
+		if configType, err := ae.GetHandlers().ConfigType.Read(val); err == nil && configType != nil {
+			result = append(result, val)
+		} else if configType, err := ae.GetHandlers().ConfigType.ReadByName(val); err == nil && configType != nil {
+			result = append(result, configType.Id)
+		}
```
dovholuknf: should a fall through branch be added 'just in case'?

controller/internal/routes/authenticate_router.go
plorenz: Probably handle that with some sort of layer of indirection, so that we use some special token under the hood, but allow different types to be mapped to it.

The prefixing could also be tidied at some point, so that if we want to change the syntax we don't have to do as much hunting.

controller/internal/routes/authenticate_router.go
plorenz: Could add a debug log statement

controller/model/config_type_model.go
```diff
@@ -0,0 +1,62 @@
...
+
+type ConfigType struct {
+	BaseModelEntityImpl
+	Name string
+	Type string
+	Data map[string]interface{}
+}
+
+func (entity *ConfigType) ToBoltEntityForCreate(_ *bbolt.Tx, _ Handler) (persistence.BaseEdgeEntity, error) {
+	if entity.Name == CONFIG_TYPE_ALL {
```
dovholuknf: is this input trimmed before being compared?

plorenz: No, it's whatever is in the data store. We might want to put something in the backlog to consider if we want to handle trimming in the API or model layers. 

This will at least be handled consistently in that it won't be compared to trimmed values later.

Issue #43, opened 2020-01-16
Update policies when entities related by name change


plorenz:
Ensure the @<name> is updated when the entity name is updated.

PR #53, opened 2020-01-21
Update policies when entities referenced by name change their names. …

plorenz:
…Fixes GH-43

### DIFF
--- a/controller/persistence/base_store.go
+++ b/controller/persistence/base_store.go
@@ -108,6 +108,48 @@ func (store *baseStore) baseLoadOneById(tx *bbolt.Tx, id string, entity boltz.Ba
 	return nil
 }
 
+func (store *baseStore) updateEntityNameReferences(bucket *boltz.TypedBucket, rolesSymbol boltz.EntitySetSymbol, entity NamedEdgeEntity, oldName string) {
+	// If the entity name is the same as entity ID, bail out. We don't want to remove any references by ID, since
+	// those take precedence over named references
+	if store.IsEntityPresent(bucket.Tx(), oldName) {
+		return
+	}
+	oldNameRef := entityRef(oldName)
+	newNameRef := entityRef(entity.GetName())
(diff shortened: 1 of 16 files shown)


controller/persistence/base_store.go
```diff
@@ -108,6 +108,48 @@ func (store *baseStore) baseLoadOneById(tx *bbolt.Tx, id string, entity boltz.Ba
 	return nil
 }
 
+func (store *baseStore) updateEntityNameReferences(bucket *boltz.TypedBucket, rolesSymbol boltz.EntitySetSymbol, entity NamedEdgeEntity, oldName string) {
+	// If the entity name is the same as entity ID, bail out. We don't want to remove any references by ID, since
+	// those take precedence over named references
+	if store.IsEntityPresent(bucket.Tx(), oldName) {
+		return
```
dovholuknf: i feel like this needs a log, or to return an err maybe? If not - why not?

controller/persistence/base_store.go
```diff
@@ -108,6 +108,48 @@ func (store *baseStore) baseLoadOneById(tx *bbolt.Tx, id string, entity boltz.Ba
...
+		})
+		if err != nil {
+			bucket.SetError(err)
+			return
+		}
+	}
+}
+
+func (store *baseStore) deleteEntityReferences(tx *bbolt.Tx, entity NamedEdgeEntity, rolesSymbol boltz.EntitySetSymbol) error {
+	// If the entity name is the same as entity ID, don't remove any of those references as id references take precedence
```
dovholuknf: same comment as above. seems like a really hard situation to detect in the off chance this happens?

controller/persistence/base_store.go
plorenz: I'll add a log statement

controller/persistence/base_store.go
plorenz: I'll add a log statement

PR #57, opened 2020-01-22
Allow anyOf semantics in addition to allOf for policies. Fixes GH-41

plorenz:


controller/persistence/base_entity.go
```diff
@@ -44,12 +44,18 @@ const (
...
 	FieldRoleAttributes = "roleAttributes"
 	FieldCreatedAt      = "createdAt"
 	FieldUpdatedAt      = "updatedAt"
 	FieldTags           = "tags"
+
+	SemanticAllOf = "AllOf"
+	SemanticAnyOf = "AnyOf"
 )
 
+var validSemantics = []string{SemanticAllOf, SemanticAnyOf}
```
andrewpmartinez: Seems weird this is here, but it doesn't hurt anything.

controller/persistence/edge_router_policy_store_test.go
```diff
@@ -51,6 +51,10 @@ func (ctx *TestContext) testEdgeRouterPolicyInvalidValues(_ *testing.T) {
 	err := ctx.create(policy)
 	ctx.EqualError(err, fmt.Sprintf("the value '[%v]' for 'identityRoles' is invalid: no identities found with the given names/ids", invalidId))
 
+	policy.IdentityRoles = []string{AllRole, roleRef("other")}
+	err = ctx.create(policy)
+	ctx.EqualError(err, fmt.Sprintf("the value '[%v %v]' for 'identityRoles' is invalid: if using %v, it should be the only role specified", AllRole, roleRef("other"), AllRole))
```
andrewpmartinez: As a heads up `t *Testing.T` is unique per `t.Run()` -  each `t.Run()` is run in its own go routine and the `t` has go routine sensitive state. `ctx` fields that use `require.Assertions` from `require.New(t)` should not be shared between different levels of `t.Run()`. Doing so will cause a panic error that hides the test results diffs/error messages/etc.

It isn't pretty but for each `t.Run()` if `require` is going to be use I would expect to see a new `require.New(t)`.

controller/persistence/base_entity.go
plorenz: It's shared across multiple entity types, so I put it there b/c I couldn't think of a better place.

controller/persistence/edge_router_policy_store_test.go
plorenz: Adding the following in a follow up PR

```
func (ctx *TestContext) testContextChanged(t *testing.T) {
       ctx.req = require.New(t)
}
```

Issue #44, opened 2020-01-16
Edge API consistency


ekoby:
in new service model services have permissions:
`permissions: [Bind,Dial]`.
to request new session we should use same terminology, like this:
`{ serviceId: 000-00-00, type: Dial }`

instead of `hosting: true`

PR #60, opened 2020-01-22
Replacing isHosting with session type. Fixes GH-44

plorenz:


pb/edge_ctrl_pb/edge_ctrl.pb.go
```diff
@@ -1,19 +1,3 @@
-/*
```
dovholuknf: i wonder what 'the policy' is for generated code... i've never researched it. ¯\_(ツ)_/¯

plorenz: I didn't even notice. In part b/c github keeps these collapsed by default.

PR #61, opened 2020-01-23
Move tunneler configuration into config entities

plorenz:


controller/internal/routes/config_type_router.go
```diff
@@ -17,9 +17,11 @@
 package routes
```
plorenz: Adding list of configs of this config type

controller/internal/routes/service_router.go
```diff
@@ -64,6 +64,11 @@ func (ir *ServiceRouter) Register(ae *env.AppEnv) {
 
 	sr.HandleFunc(servicePolicyUrl, servicePoliciesListHandler).Methods(http.MethodGet)
 	sr.HandleFunc(servicePolicyUrl+"/", servicePoliciesListHandler).Methods(http.MethodGet)
```
plorenz: Adding service to list configs associated with service

controller/persistence/migration_v4.go
```diff
@@ -0,0 +1,67 @@
+/*
```
plorenz: Creates tunneler client/server config types. Migrates dns configuration into config entities

tunnel/intercept/svcpoll.go
```diff
@@ -105,47 +109,65 @@ func diffServices(edgeServices []edge.Service, knownServices map[string]edge.Ser
...
-			} else {
-				log.Warnf("service %v is hostable but is missing a dial address. Add a 'tunneler.dial.addr' tag to the service to fix", svc.Name)
+				go host(context, svc)
+			} else if !found {
+				log.WithError(err).Warnf("service %v is hostable but no server config of type %v is available", svc.Name, entities.ServerConfigV1)
+			} else if err != nil {
+				log.WithError(err).Errorf("service %v is hostable but unable to decode server config of type %v", svc.Name, entities.ServerConfigV1)
 			}
 		}
 	}
```
plorenz: We don't know here if the intercept above actually worked. Should we do anything different if it failed (if, for example, the config was missing or invalid)?

tunnel/intercept/svcpoll.go
```diff
@@ -105,47 +109,65 @@ func diffServices(edgeServices []edge.Service, knownServices map[string]edge.Ser
 	return
 }
 
-func updateServices(context ziti.Context, interceptor Interceptor, resolver dns.Resolver, added, removed, all map[string]edge.Service) {
+func updateServices(context ziti.Context, interceptor Interceptor, resolver dns.Resolver, added, removed, all map[string]*entities.Service) {
 	log := pfxlog.Logger()
 	for _, svc := range added {
 		if stringz.Contains(svc.Permissions, "Dial") {
```
plorenz: We don't fail here if no config is found because proxy works fine without configuration

tunnel/intercept/svcpoll.go
```diff
@@ -105,47 +109,65 @@ func diffServices(edgeServices []edge.Service, knownServices map[string]edge.Ser
...
+			clientConfig := &entities.ServiceConfig{}
+			if found, err := svc.GetConfigOfType(entities.ClientConfigV1, clientConfig); found && err == nil {
+				svc.ClientConfig = clientConfig
+			} else if !found {
```
scareything: I wonder if these messages will make folks think something is broken when it really isn't?

For the `!found` case, would it be reasonable to log at Debug here and let the interceptor log an error as appropriate?

For `found && err !=nil`, it seems like this should be logged as Error and the loop immediately continues?

plorenz: Makes sense. If it can't find the client config, should we let it try to host it?

scareything: Oh, I had to think about why you asked this question... I was originally thinking that the configs were all grabbed at once, but the server config is grabbed separately below/. So you're saying that an error here reading the client config shouldn't continue the loop. Agreed.

tunnel/intercept/svcpoll.go
scareything: Good question. As it is right now the address of a service with client config that fails to intercept will be added to the "all" map. It isn't ideal but I think it's acceptable, since the only harm that will come from it is a doomed call to StopIntercepting.

Avoiding this would require keeping track of services that we are aware of but not intercepting (lest we attempt to set up the intercept on every poll cycle).

tests/context.go
```diff
@@ -94,6 +94,10 @@ func GetTestContext() *TestContext {
 	return defaultTestContext
 }
 
+func (ctx *TestContext) testContextChanged(t *testing.T) {
+	ctx.req = require.New(t)
+}
+
```
andrewpmartinez: I want to noodle on what this section of code addresses. I don't like the idea of even temporarily sharing `t *testing.T` and having a global that is set and shared across tests limits us being able to have to have parallel testing (should our tests start to take too long). 

I had some POC code that I didn't like that allows a `testContext` to generate a child `testContext`. I think the idea has legs, but not that quick and dirty pass.

Leave this for now I suppose and we can work to improve it as time goes on.

plorenz: Ok, makes sense. Looking forward to it :)

PR #63, opened 2020-01-24
Enable schema validation for configs. Fixes GH-62

plorenz:


controller/embedded/api-schema/entities/config-type.post.json
```diff
@@ -9,6 +9,13 @@
     "name": {
       "$ref": "definitions.all.json#/definitions/name"
     },
+    "schema": {
+      "type": [
+        "object",
+        "null"
+      ],
+      "default": null
```
andrewpmartinez: IIRC the JSON schema parser used in GO doesn't actually use these values. The Node.js one did.

They don't hurt and are part of the spec. Most of the time the GO JSON parser defaults to an empty value such as NIL but other times won't if the empty value is a non-nil value ("", 0, etc)

plorenz: I've noticed that with the API schemas, that I had to manage defaults manually in the code. 

If we start to care about it more, we can probably put together a PR to fix in the library.

tests/config_test.go
```diff
@@ -232,6 +233,45 @@ func Test_Configs(t *testing.T) {
 		ctx.validateDateFieldsForCreate(now, entityJson)
 
 		ctx.AdminSession.requireDeleteEntity(config)
-		ctx.requireNotFoundError(ctx.AdminSession.query("configs/" + uuid.New().String()))
+		ctx.requireNotFoundError(ctx.AdminSession.query("configs/" + config.id))
+	})
+
+	ctx.enabledJsonLogging = true
```
andrewpmartinez: Expected?

tests/config_type_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+	t.Run("update config type should pass", func(t *testing.T) {
+		ctx.testContextChanged(t)
+		now := time.Now()
+		configType := ctx.AdminSession.requireCreateNewConfigType()
+
+		entityJson := ctx.AdminSession.validateEntityWithQuery(configType)
+		createdAt := ctx.validateDateFieldsForCreate(now, entityJson)
+
+		time.Sleep(time.Millisecond * 10)
```
andrewpmartinez: What behaviour were you seeing that required `time.Sleep()`?

controller/embedded/api-schema/entities/config-type.post.json
andrewpmartinez: I vaguely recall the author saying "no" to a direct ask for that.

tests/config_test.go
plorenz: Missed that. I've removed it for the next pass.

tests/config_type_test.go
plorenz: the validateDateFieldsForUpdate below checks that the updateAt field has been updated since the create above, so it needs a small gap to ensure that it actually does increase.

controller/embedded/api-schema/entities/config-type.post.json
plorenz: I found it: https://github.com/xeipuuv/gojsonschema/issues/45 (with your comment :) )

Sounds like the maintainer didn't have time to pull that PR into shape, but would be ok with someone else doing it.

andrewpmartinez: So not a "no" just a "not me".

PR #68, opened 2020-01-31
Allow overriding service configurations per identity

plorenz:


controller/model/base_model.go
```diff
@@ -32,8 +32,15 @@ type BaseModelEntity interface {
 	GetCreatedAt() time.Time
 	GetUpdatedAt() time.Time
 	GetTags() map[string]interface{}
+}
+
+type BoltEntitySink interface {
```
ekoby: are we leaking persistence implementation details?

plorenz: The model needs to know how to convert to/from bolt types. However, I should probably make these types package private, since nothing outside the model package needs them.

plorenz: Fixed

PR #72, opened 2020-02-03
Extract service edge router policies into standalone entity. Fixes GH-69

plorenz:


controller/embedded/api-schema/entities/service-edge-router-policy.post.json
```diff
@@ -0,0 +1,29 @@
...
+  "required": [
+    "name"
+  ],
+  "properties": {
+    "name": {
+      "$ref": "definitions.all.json#/definitions/name"
+    },
+    "semantic": {
+      "default": null,
+      "$ref": "definitions.all.json#/definitions/semantic"
```
andrewpmartinez: Even though `default` here is impotent, would null be a valid value for semantic? I assume it defaults to the `all` semantic.

plorenz: Yes, if default worked, the default would be all. But since it doesn't, it's just trying to indicate that you can pass in nothing or null and then the model code will default it to any for you.

controller/internal/routes/edge_router_router.go
```diff
@@ -47,11 +47,11 @@ func (ir *EdgeRouterRouter) Register(ae *env.AppEnv) {
 	sr := registerCrudRouter(ae, ae.RootRouter, ir.BasePath, ir, permissions.IsAdmin())
 	registerCrudRouter(ae, ae.RootRouter, ir.BasePathLegacy, ir, permissions.IsAdmin())
 
-	servicesUrl := fmt.Sprintf("/{%s}/%s", response.IdPropertyName, EntityNameService)
-	servicesListHandler := ae.WrapHandler(ir.ListServices, permissions.IsAdmin())
+	serviceEdgeRouterPoliciesUrl := fmt.Sprintf("/{%s}/%s", response.IdPropertyName, EntityNameServiceEdgeRouterPolicy)
+	serviceEdgeRouterPoliciesListHandler := ae.WrapHandler(ir.ListServiceEdgeRouterPolicies, permissions.IsAdmin())
 
-	sr.HandleFunc(servicesUrl, servicesListHandler).Methods(http.MethodGet)
```
andrewpmartinez: Providing an endpoint of effective services is still possible. This change makes it so that we expose the internal model of ER->SERPs and that is fine.

We will want those "effective" endpoints in the future for security auditing.

controller/model/edge_router_handlers_test.go
```diff
@@ -25,18 +25,17 @@ func (ctx *TestContext) testGetEdgeRoutersForServiceAndIdentity(*testing.T) {
 	ctx.requireNewEdgeRouterPolicy(ss("#all"), ss("#all"))
 
 	// test default case, with no limits on service
-	ctx.True(ctx.isEdgeRouterAccessible(edgeRouter.Id, identity.Id, service.Id))
-	ctx.True(ctx.isEdgeRouterAccessible(edgeRouter2.Id, identity.Id, service.Id))
+	ctx.False(ctx.isEdgeRouterAccessible(edgeRouter.Id, identity.Id, service.Id))
```
andrewpmartinez: Why would these switch to false?

controller/persistence/cluster_store.go
```diff
@@ -46,6 +46,7 @@ func (entity *Cluster) GetEntityType() string {
 type ClusterStore interface {
 	Store
```
andrewpmartinez: Is this still used?

controller/internal/routes/edge_router_router.go
plorenz: Agree. I know we're going to have to do another pass where we add a bunch more endpoints for UI awesomeness.

controller/model/edge_router_handlers_test.go
plorenz: In the old model if you didn't specify edgeRouterRoles on the service, the service had access to all edge routers. 
In the new model a service needs to be added to at least one service edge router policy in order to have access to any edge routers

controller/persistence/cluster_store.go
plorenz: Only by migration code. I've been trying to reduce the entangling of the deprecated stores, but there's more I can do. Also, I figure we can strip everything out in Ziti 1.1.

Issue #83, opened 2020-02-14
Failure returned when listing services for a user with no access to services, but with services in the system


plorenz:
```
error: error listing https://localhost:1280/services?asIdentity=all in Ziti Edge Controller. Status code: 500 Internal Server Error, Server returned: {"error":{"args":{"cause":{},"urlVars":{}},"cause":{},"causeMessage":"identity with all id or name not found","code":"UNHANDLED","message":"An unhandled error occurred","requestId":"3c04809a-6996-454f-b951-8dbb797bad92"},"meta":{"apiEnrolmentVersion":"0.0.1","apiVersion":"0.0.1"}}
``` 
in the tunneler

Issue #76, opened 2020-02-10
lookup router via id or name


dovholuknf:
right now the router id must be used when creating services. it'd be more friendly to use the provided input as an id first and if not found by id then to look up a router with the same name. if only one router matches use the id of the provided router and log output stating that it was found by name not id

smilindave26:
Clint - in my experience, APIs that "guess" at what the user intended often do more harm than good.  I think its better to just require the id and leave it at that.

dovholuknf:
Generally speaking I agree with your point of view but it makes using the ziti cli "manually" a pain. We also already have this pattern in place for other entities - we could remove the functionality from the other locations too but in my opinion it really is easier on the user if it works this way. it's also not guessing about the name as much as falling back to using the name if the id isn't matched.

@plorenz FYI

plorenz:
> Clint - in my experience, APIs that "guess" at what the user intended often do more harm than good. I think its better to just require the id and leave it at that.

@smilindave26 We do this in lots of other places, when referencing services and identities. I don't see the problem doing this for edge routers as well, especially since the router ids are UUIDs. Router names may be UUIDs or may be whatever else, but either way, the only way there would be a collision between a router id and name is if someone did it intentionally. 

If you feel strongly about it, we don't have to do it, but then we should probably considered stripping it out everywhere else

smilindave26:
No, don't strip it out and go ahead and stay consistent here...

PR #84, opened 2020-02-14
* Fix bug where overriden query was ignored. Fixes GH-83

plorenz:
* Allow routers to referenced by name when creating services. Fixes GH-76

* Allow admins to view services as other users

* Allow anyone to change the requested config types shown in a services list

### DIFF
--- a/controller/internal/routes/service_router.go
+++ b/controller/internal/routes/service_router.go
@@ -18,7 +18,9 @@ package routes
 
 import (
 	"fmt"
+	"github.com/netfoundry/ziti-edge/controller/util"
 	"net/http"
+	"strings"
 
 	"github.com/michaelquigley/pfxlog"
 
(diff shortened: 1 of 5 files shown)


controller/internal/routes/service_router.go
```diff
@@ -74,7 +76,27 @@ func (ir *ServiceRouter) Register(ae *env.AppEnv) {
...
+			}
+		}
+
+		// allow overriding config types
+		configTypes := rc.ApiSession.ConfigTypes
+		if requestedConfigTypes := rc.Request.URL.Query().Get("configTypes"); requestedConfigTypes != "" {
+			configTypes = mapConfigTypeNamesToIds(ae, strings.Split(requestedConfigTypes, ","), identity.Id)
+		}
+
+		result, err := ae.Handlers.Service.PublicQueryForIdentity(identity, configTypes, queryOptions)
```
andrewpmartinez: Been wondering what `PublicQuery` means.

plorenz: PublicQuery is mean to be called for queries which come from the APIs. It checks that queries are using only local symbols and aren't chaining symbols. 

So for example when querying service, `name in ["foo", "bar"], createdAt > someDate` is ok, but `sessions.identity.name = "Admin"` would result in an error. 

Internal users would use just Query and wouldn't restricted in the symbols available to them

controller/internal/routes/service_router.go
```diff
@@ -74,7 +76,27 @@ func (ir *ServiceRouter) Register(ae *env.AppEnv) {
...
+				var err error
+				identity, err = ae.Handlers.Identity.ReadOneByQuery(fmt.Sprintf(`id = "%v" or name = "%v"`, asId, asId))
+				if err != nil {
+					return nil, err
+				}
+				if identity == nil {
+					return nil, util.NewNotFoundError("identity", "id or name", asId)
+				}
+			}
+		}
```
plorenz: @dovholuknf @mikegorman-nf implementation code

PR #88, opened 2020-02-25
add password for self service password, fixes #87

andrewpmartinez:
- moves logic from REST endpoint def to handlers
- moves to version 0.10
- adds tests for invalid current passwords
- sync password length mins between init and API


controller/internal/routes/base_router.go
```diff
@@ -197,8 +197,8 @@ func Create(rc *response.RequestContext, rr response.RequestResponder, sc *gojso
 		rr.RespondWithCouldNotParseBody(err)
 		return
 	}
-
-	il := gojsonschema.NewGoLoader(in)
+	
+	il := gojsonschema.NewBytesLoader(body)
```
andrewpmartinez: This is the validation change from go object to JSON I mentioned in the call. Happens 3 times (create, update, patch)

controller/internal/routes/current_identity_authenticator_router.go
```diff
@@ -86,69 +84,19 @@ func (ir *CurrentIdentityAuthenticatorRouter) Detail(ae *env.AppEnv, rc *respons
...
-
-		if authenticator == nil {
-			return util.NewNotFoundError(ae.GetHandlers().Authenticator.GetStore().GetSingularEntityType(), "id", id)
-		}
-
-		if authenticator.IdentityId != rc.Identity.Id {
-			return apierror.NewUnhandled()
+	apiEntity := &AuthenticatorSelfUpdateApi{}
+	Patch(rc, ae.Schemes.AuthenticatorSelf.Patch, ir.IdType, apiEntity, func(id string, fields JsonFields) error {
+		if fields.IsUpdated("newPassword") {
```
andrewpmartinez: @plorenz there I had to add this as there is a computed field (salt) and fields in the API that don't match the model/persistence field names.

I don't know if there is a better way, but this was "easy." I wanted to do it in the handlers (so that any presentation would have the logic), but I didn't know if extending the boltz checker interface was the right call as it could be a conjunction of the handler and json fields.

plorenz: There's an AndFieldChecker which combines two checkers. You could do a similar thing with an OrFieldChecker, which would be pretty simple.

However, this isn't offensive either :)

andrewpmartinez: I want it in the handler - any logic outside the handlers means it has to be duplicated for other presentations.

I've added the disjunction field checker.

Issue #92, opened 2020-03-02
Add API to list role attributes in use by services, identities and edge routers


plorenz:
Need the following APIs

* `/services/role-attributes`
* `/identities/role-attributes`
* `/edge-routers/role-attributes`

Each API service should return a unique array of role attributes associated with the given entity

plorenz:
Going to go with 

* /role-attributes-services
* /role-attributes-identities
* /role-attributes-edgei-routers

PR #103, opened 2020-03-22
Implement role attribute querying. Fixes GH-92

plorenz:


controller/env/appenv.go
```diff
@@ -393,6 +393,12 @@ func (ae *AppEnv) WrapMiddleware(f AppMiddleware) mux.MiddlewareFunc {
 	}
 }
 
+func (ae *AppEnv) HandleGet(path string, f AppHandler, prs ...permissions.Resolver) {
+	handler := ae.WrapHandler(f, prs...)
+	ae.RootRouter.HandleFunc(path, handler).Methods(http.MethodGet)
```
dovholuknf: how is this not a utility function yet? :)

dovholuknf: Also - going forward how do we keep the openapi spec up to date? @andrewpmartinez will the spec generate the server code for us? so that we'd update the spec/server code first?

plorenz: Just in time for it be made irrelevant by the swagger work :) (maybe, I haven't dug into it much yet).

go.mod
```diff
@@ -2,9 +2,9 @@ module github.com/netfoundry/ziti-edge
 
 go 1.13
 
-// replace github.com/netfoundry/ziti-foundation => ../ziti-foundation
+replace github.com/netfoundry/ziti-foundation => ../ziti-foundation
```
dovholuknf: re-comment?

controller/env/appenv.go
andrewpmartinez: When I get my worked merged the workflow will be:

1) write swagger
2) generate server code
3) wire business logic in handlers to REST API presentation.

Right now it is getting a little stale, but the goal is to get the new workflow out this week.

PR #105, opened 2020-03-24
Auto Provision Identity Attributes From CA

andrewpmartinez:


controller/embedded/api-schema/entities/ca.put.json
```diff
@@ -25,6 +25,10 @@
       "type": "boolean",
       "default": false
     },
+    "identityRoles": {
```
andrewpmartinez: @plorenz I believe this is a good attribute name but let me know if not.

tests/ca_test.go
```diff
@@ -0,0 +1,79 @@
+// +build apitests
```
andrewpmartinez: @plorenz there are some testing facilities that you have built that I am not sure I am taking advantage of. Let me know if anything jumps out.

tests/entities.go
```diff
@@ -557,3 +566,179 @@ func (entity *configValidatingService) validate(ctx *TestContext, c *gabs.Contai
...
+	tags                      map[string]interface{} `json:"tags"`
+
+	privateKey crypto.Signer     `json:"-"` //utility property, not used in API calls
+	publicCert *x509.Certificate `json:"-"` //utility property, not used in API calls
+}
+
+func newTestCa(identityRoles ...string) *ca {
+	key, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		panic(err)
```
andrewpmartinez: Not sure what to do w/ these types of errors. They *should* never happen.

controller/embedded/api-schema/entities/ca.put.json
plorenz: Makes sense to me

tests/entities.go
plorenz: I think for those kinds of errors, panic is fine. I would probably pass the test context in and use require.NoError, but only if it was convenient to do so

controller/embedded/api-schema/entities/ca.put.json
ekoby: `defaultIdentityRoles` ?

Issue #99, opened 2020-03-20
Add policy advisor API


plorenz:
API should be able help diagnose policy problems. 

* iterate across all identities. See if they have access to services but no routers for those services.
* Given a specific identity and service, show if they should be able to dial/bind it, and if not, why not.
* Iterate across routers, and ensure that something is able to use them. 
* Given two entities that are linked, show via which policies they are linked

PR #120, opened 2020-03-31
Initial policy advisor API. Fixes GH-99

plorenz:


controller/internal/routes/identity_api_model.go
```diff
@@ -317,3 +317,46 @@ type IdentityServiceConfig struct {
 func (entity IdentityServiceConfig) toModel() model.ServiceConfig {
 	return model.ServiceConfig{Config: entity.Config, Service: entity.Service}
 }
+
+type AdvisorEdgeRouter struct {
+	*EntityApiRef
+	Online bool `json:"isOnline"`
```
andrewpmartinez, line 323: I am partial to `IsOnline` for the struct property name as well.

controller/persistence/service_policy_store.go
```diff
@@ -92,6 +92,13 @@ func (entity *ServicePolicy) GetEntityType() string {
...
+func (entity *ServicePolicy) GetPolicyTypeName() string {
+	if entity.PolicyType == PolicyTypeBind {
+		return PolicyTypeBindName
+	}
+	return PolicyTypeDialName
+}
+
 type ServicePolicyStore interface {
 	NameIndexedStore
 	LoadOneById(tx *bbolt.Tx, id string) (*ServicePolicy, error)
```
andrewpmartinez, line 104: I didn't see any API tests

specs/swagger.yml
```diff
@@ -1264,6 +1266,25 @@ paths:
...
+      - $ref: '#/parameters/id'
+    get:
+      summary: List associated identities
+      description: |
+        Retrieves a list of identities that may access services via the given edge router. Supports filtering, sorting, and pagination. Requires admin access.
+      security:
+        - ztSession: []
+      tags:
+        - Edge Router
+      operationId: listEdgeRoutersIdentities
```
andrewpmartinez, line 1280: `listEdgeRouterIdentities`   less 1 `s`

specs/swagger.yml
```diff
@@ -1282,6 +1303,26 @@ paths:
...
+      - $ref: '#/parameters/id'
+    get:
+      summary: List associated services
+      description: |
+        Retrieves a list of services that may be accessed via the given edge router. Supports filtering, sorting, and pagination. Requires admin access.
+      security:
+        - ztSession: []
+      tags:
+        - Edge Router
+      operationId: listEdgeRoutersServices
```
andrewpmartinez, line 1318: Less 1 `s`, `EdgeRouter`

specs/swagger.yml
```diff
@@ -1689,6 +1732,69 @@ paths:
...
+        '401':
+          $ref: '#/responses/unauthorizedResponse'
+
+  '/identities/{id}/policy-advice/{serviceId}':
+    parameters:
+      - $ref: '#/parameters/id'
+      - $ref: '#/parameters/serviceId'
+    get:
+      summary: Analyze policies relating the given identity and service
+      description: |
```
andrewpmartinez, line 1781: I find this site helpful when dealing with long strings in YAML. YAML is much better at them than JSON.

https://yaml-multiline.info/

Lots of little knobs to turn.

specs/swagger.yml
```diff
@@ -2962,6 +3195,22 @@ responses:
     description: Returns the document that represents the specification
     schema:
       $ref: '#/definitions/detailSpecBodyEnvelope'
+  ###################################################################
+  # Role Attributes
+  ##################################################################
+  listRoleAttributes:
+    description: A list of specifications
```
andrewpmartinez, line 3202: Copy pasta: `specifications`

specs/swagger.yml
```diff
@@ -2962,6 +3195,22 @@ responses:
...
+    schema:
+      $ref: '#/definitions/listRoleAttributesEnvelope'
+
+  ###################################################################
+  # Policy Advisor
+  ##################################################################
+  getIdentityPolicyAdvice:
+    description: Returns the document that represents the specification
```
andrewpmartinez, line 3210: Copy pasta: `specifications`

specs/swagger.yml
```diff
@@ -4802,3 +5051,80 @@ definitions:
...
+      - "identityId"
+      - "identity"
+      - "serviceId"
+      - "service"
+      - "isBindAllowed"
+      - "isDialAllowed"
+      - "identityRouterCount"
+      - "serviceRouterCount"
+      - "commonRouters"
+    properties:
```
andrewpmartinez, line 5089: `required` section not necessary as this is a read object.

specs/swagger.yml
```diff
@@ -4802,3 +5051,80 @@ definitions:
...
+        items:
+          type: "object"
+          properties:
+            entity:
+              type: "string"
+            id:
+              type: "string"
+            name:
+              type: "string"
+            _links:
```
andrewpmartinez, line 5117: `_links` should be 

```      _links:
        $ref: '#/definitions/links'```

File: specs/swagger.yml [hunk 10]
```diff
@@ -4802,3 +5051,80 @@ definitions:
...
+      - "serviceId"
+      - "service"
+      - "isBindAllowed"
+      - "isDialAllowed"
+      - "identityRouterCount"
+      - "serviceRouterCount"
+      - "commonRouters"
+    properties:
+      identityId:
+        type: "string"
```
andrewpmartinez, line 5091: With `id` values we should just go all string.

If you are in this file again I would just find: `format: uuid` and remove them all. We have to be 100% consistent otherwise consumers of this will need to convert string <-> uuid.

andrewpmartinez:
Also, the swagger file is now embedded into the controller so it can be served on `GET /specs/swagger/spec`. `packr -z` needs to be run.

controller/internal/routes/identity_api_model.go
plorenz: Fixed

specs/swagger.yml
plorenz: Fixed

specs/swagger.yml
plorenz: Fixed

specs/swagger.yml
plorenz: Fixed

specs/swagger.yml
plorenz: Fixed

specs/swagger.yml
plorenz: Removed

specs/swagger.yml
plorenz: They are entityRefs extended by the IsOnline field, so didn't think I could use entityRef

specs/swagger.yml
plorenz: Done. I remove `format: uuid` everywhere it was used by ids, but I left other uses, such as for tokens, etc

controller/persistence/service_policy_store.go
plorenz: Yes, I weighed writing test vs getting back to HA features and skipped them for now. I think we want them in the medium term, but in the short term, since this is advisor code, I thought it was okay.

plorenz:
> See comments

I believe all comments have been addressed, please take another look

specs/swagger.yml
andrewpmartinez: oof. I didn't see isOnline there ate the bottom.

Is that necessary? The way this will work w/ generators will be an anonymous struct/class will be generated. Depending on the language this may or may not be great.

plorenz: I need the IsOnline associated with the router. I could structure it differently, so that we've got

```
commonRouters : [
    {
        isOnline: true,
        router: {
            // entityRef here
        }
    }
]
```
Which feels weird. Could I just extract out that definition as CommonRouterEntityRef or something like that so that it's named?

andrewpmartinez: You can define a generic map of additional attributes.

We would update the entityRef definition to had the following property:

```
  entityRef:
    //...existing def
    properties:
      //exisiting def
      additionalPropertiesThatYouMayCareAbout:
        type: object
        properties:
          isOnline:
            type: boolean
          someOtherProperty:
            type: string
               
```

andrewpmartinez: Thinking about this more, we can just add a generic data map to entityRef:



```
entityRef:
  //existing
  properties:
    //existing
    data:
      type: object
      additionalProperties:
        $ref: '#/definitions/dataOptions'
```

PR #123, opened 2020-04-01
Tx.router.extended.base

### LINKED ISSUE #27: Add Transit Router Enrollment

andrewpmartinez:
- consolidates enrollment logic to use the same pattern (i.e. edge routers are no longer unique)
- adds transit router support to the Edge API
- adds transit router enrollment to the Edge API, resolves #27 
- updates Edge API tests to inspect and validate CRUD responses for proper envelope format/parsing
- adds transit router create/enroll/delete tests


controller/persistence/migrations.go
```diff
@@ -63,10 +63,14 @@ func (m *Migrations) migrate(step *boltz.MigrationStep) int {
 		return 4
 	}
 
-	// current version
 	if step.CurrentVersion == 4 {
```
plorenz, line 66: This should return 5

controller/persistence/transit_router_store.go
```diff
@@ -0,0 +1,204 @@
...
+
+	store.indexName = store.addUniqueNameField()
+	store.symbolEnrollments = store.AddFkSetSymbol(FieldTransitRouterEnrollments, store.stores.enrollment)
+}
+
+func (store *transitRouterStoreImpl) initializeLinked() {
+
+	store.AddLinkCollection(store.symbolEnrollments, store.stores.enrollment.symbolIdentity)
+
+	//todo: confirm this is needed
```
plorenz, line 130: I don't think this is needed. Should delete the base store bucket which will clear everything underneath.

controller/persistence/migrations.go
andrewpmartinez: Fixed

controller/persistence/transit_router_store.go
andrewpmartinez: Fixed

PR #128, opened 2020-04-03
Rest.client.server.gen

andrewpmartinez:
All of the files in /rest_server and /rest_client are 100%  generated from the swagger files.

The helper scripts in /scripts invoke `swagger` (from go-swagger) to generate the client and server.

This PR does not contain any of the logic to wire the generated client or server to any existing tests, CLIs, etc.  That will be in upcoming PRs. The separation is to make this somewhat bearable. 

Commit message:

- bash/powershell, mitigates OS output differences for easier PRs
- inserts copyright information
- script logic is location-sensitive, running it from anywhere
  work and put the resulting generated code in the repo where
  the script is located.
  example: /home/me/repos/netfoundry/ziti-edge/scripts/generate_rest.sh
- fix CA naming


models/service_detail.go
```diff
@@ -0,0 +1,246 @@
...
+
+import (
+	"strconv"
+
+	"github.com/go-openapi/errors"
+	"github.com/go-openapi/strfmt"
+	"github.com/go-openapi/swag"
+	"github.com/go-openapi/validate"
+)
+
```
plorenz, line 40: Service Detail doesn't have the `config` attribute with the actual embedded configuration data for the config types the user expressed interest in. Does that need to be added or are you planning a different API to get that data?

andrewpmartinez: Switched back to array of strings.

plorenz: Ah, so you saw a different incompatibility than I did. I think change from []string -> []*EntityRef for configs is fine and an improvement, even it's not backwards compatible. I think we should probably actually remove configs all together, since you can get them using /services/<id>/configs. 

What I was concerned about was config, which embeds the actual config data related to a service in the service definition. So you'll see something like this:
```
            "name": "iperf",
            "terminatorStrategy": "",
            "roleAttributes": null,
            "permissions": [
                "Bind",
                "Dial"
            ],
            "configs": [
                "1a6290b4-39fb-4bf0-9eac-3e3fb2906029"
            ],
            "config": {
                "ziti-tunneler-client.v1": {
                    "hostname": "iperf",
                    "port": 5201
                }
            }
```
That still needs to be there, b/c non-admins need to be able to get the config related to the service.

PR #142, opened 2020-04-27
Update for Xt changes

plorenz:


controller/embedded/api-schema/entities/terminator.put.json
```diff
@@ -20,6 +20,12 @@
     "address": {
       "type": "string"
     },
+    "cost": {
+      "type": [
+        "integer",
```
andrewpmartinez, line 25: Are there valid values for this?

IIRC JSON parsing will provide an `float64` unless the `Number` type is used, we force it to `int` and then `uint16`. If I am reading this right, I can put values in here that will default to 0.

Might want to at least include `"minimum": 0` and `"maximum": 65535`

specs/swagger.yml
```diff
@@ -5041,6 +5041,9 @@ definitions:
         type: string
       address:
         type: string
+      cost:
+        type: integer
```
andrewpmartinez, line 5045: same as above for minimum and maximum values. Swagger defaults integer to 32bits.

```
   minimum: 0
   maximum: 65535
```

andrewpmartinez: might want to define cost in 1 place under definitions and reuse.

tests/data_flow_test.go
```diff
@@ -21,6 +21,7 @@ package tests
 import (
```
andrewpmartinez, line 21: To test smart routing in data flow we will need the ability to stand up multiple edge routers right? I know the way it is currently done stores a single instance.

plorenz: Yes and now. We can test terminator selection strategies using a single router with multiple clients. I have a test written for that which I'll put up in the next pass. But for pure smart routing we'll need multiple routers since that is done just based on router <-> router metrics. I'm not sure if we need to do that in this kind of test though. Feels like that might be more appropriate for a fablab based test.

controller/embedded/api-schema/entities/terminator.put.json
plorenz: Fixed

specs/swagger.yml
plorenz: Fixed

michaelquigley:
`ziti-cmd`@`master` is broken until this is merged.

Issue #149, opened 2020-05-05
Enable graceful shutdown of bound connections


plorenz:
Allow SDK to shutdown connections gracefully, by letting them be marked as failed so open requests can be completed first without letting new ones be opened.

PR #152, opened 2020-05-07
Support graceful terminator shutdown. Add more Xt ATs. Fixes GH-149

plorenz:


gateway/xgress_edge/listener.go
```diff
@@ -296,6 +296,58 @@ func (proxy *ingressProxy) processUnbind(req *channel2.Message, ch channel2.Chan
...
+			WithField("serviceFingerprints", ns.CertFingerprints).
+			WithField("clientFingerprints", proxy.fingerprints.Prints()).
+			Error("matching fingerprint not found")
+		proxy.sendStateClosedReply("Invalid Session", req)
+		return
+	}
+
+	var cost *uint16
+	if costBytes, hasCost := req.Headers[edge.CostHeader]; hasCost {
+		updatedCost := binary.LittleEndian.Uint16(costBytes)
```
dovholuknf, line 323: should this just be a function ? `binary.LittleEndian` seems like the whole endian-ness stuff could be encapsulated. just a thought...

gateway/xgress_edge/listener.go
```diff
@@ -296,6 +296,58 @@ func (proxy *ingressProxy) processUnbind(req *channel2.Message, ch channel2.Chan
...
+
+	localListener, ok := proxy.listener.factory.hostedServices.Get(token)
+
+	if ok {
+		log.Debugf("updating terminator %v to precedence %v and cost %v", localListener.terminatorId, precedence, cost)
+		if err := xgress.UpdateTerminator(proxy.listener.factory, localListener.terminatorId, cost, precedence); err != nil {
+			log.WithError(err).Error("failed to update bind")
+		}
+	} else {
+		log.Error("failed to update bind, no listener found")
```
dovholuknf, line 347: might just be me but from just reading the code it seems like a lot of work to get here only to find the listener isn't found? probably still have to do the check for concurrency reasons but i don't remember seeing the check earlier in the function?

tests/context.go
```diff
@@ -724,42 +725,44 @@ func (server *testServer) acceptLoop() {
...
 
 	close(server.errorC)
-	fmt.Printf("%v: service exiting\n", server.idx)
+	pfxlog.Logger().Debugf("%v: service exiting", server.idx)
 }
 
 func (server *testServer) dispatch(conn *testServerConn) {
 	defer func() {
-		fmt.Printf("marking waiter done to conn %v-%v\n", conn.server.idx, conn.id)
+		// fmt.Printf("marking waiter done to conn %v-%v\n", conn.server.idx, conn.id)
```
dovholuknf, line 740: leaving this in for debugging reasons?

tests/context.go
```diff
@@ -724,42 +725,44 @@ func (server *testServer) acceptLoop() {
...
-				fmt.Printf("panic from server.dispatch: %+v\n", err)
+				log.WithError(err).Error("panic from server.dispatch")
 				server.errorC <- err
 			}
 		}
 	}()
 
 	defer func() {
-		fmt.Printf("closing conn %v-%v\n", conn.server.idx, conn.id)
+		// fmt.Printf("closing conn %v-%v\n", conn.server.idx, conn.id)
```
dovholuknf, line 757: same comment - if there's any more i won't add any more comments :)

gateway/xgress_edge/listener.go
plorenz: We do have some utility methods on request that I'd forgotten about for put/get*IntHeader. We don't have one for uint16, but I'll do that in a follow up PR and do some tidying here.

gateway/xgress_edge/listener.go
plorenz: No reason that couldn't be at the top. Moved

tests/context.go
plorenz: removed

tests/context.go
plorenz: removed

PR #155, opened 2020-05-11
Add cause to not found and debug on error return

plorenz:


controller/response/responder.go
```diff
@@ -55,6 +57,14 @@ func NewRequestResponder(rc *RequestContext) RequestResponder {
 func (rr *RequestResponderImpl) RespondWithApiError(apiError *apierror.ApiError) {
 	log := pfxlog.Logger()
 
+	if logrus.IsLevelEnabled(logrus.DebugLevel) {
+		logger := log
+		if apiError.Cause != nil {
+			logger = logger.WithError(apiError.Cause)
+		}
+		logger.Debugf("returning api error, status: %v, code: %v, msg: %v", apiError.Status, apiError.Code, apiError.Message)
```
andrewpmartinez, line 65: Is the `if` really needed? Can't we blindly call `logger.Debug` and use `stringz.OrEmpty()`?

controller/response/responder.go
```diff
@@ -150,6 +160,15 @@ func (rr *RequestResponderImpl) RespondWithNotFound() {
 	})
 }
 
+func (rr *RequestResponderImpl) RespondWithRecordNotFound(e error) {
```
andrewpmartinez, line 163: I do not see the need for both `RespondWithNotFound` and `RespondWithRecordNotFound`. If the difference is simply the `e error` used as `cause`. I'd just make `RespondWithNotFoundWithCause(e error)` and `RespondWIthNotFound()` and have one call the other.

The word "record" doesn't have any significant meaning. URLs may contain non-record accounts for virtual objects that do not have records. So I feel it muddies the waters.

controller/response/responder.go
plorenz: This if isn't needed. This just seemed like heavier than usual logging logic in what might be a busy code path. I can remove it if you feel strongly about it

Not sure where the stringz.OrEmpty comes in...?

controller/response/responder.go
plorenz: Makes sense. Fixed

controller/response/responder.go
andrewpmartinez: mmm mistakenly thought .Cause was in `*string` form here.

PR #188, opened 2020-06-05
Swagger.code.gen

andrewpmartinez:
This PR only has the most recent `swagger.yml` in the `specs/` directory, the scripts used to smooth the code-gen process in `scripts/` and the output of the code gen.

This PR changes 0 functionally, just adds a ton of unused code that won't be used till the other branch is merged. Github isn't going to be the easiest place do review the files - I'd suggest using Goland to compare master to this branch.

If you would like from within `edge/` you can run `./scripts/generate_rest.sh` to confirm the output is the same as that generated by go-swagger.


specs/swagger.yml
```diff
@@ -1382,11 +1440,14 @@ paths:
       summary: Enroll an identity vvia one-time-token
       description: |
         Enrolls and identity via a one-time-token to establish an initial username and password combination.
```
plorenz, line 1442: spelling

specs/swagger.yml
```diff
@@ -1911,7 +1973,7 @@ paths:
           required: true
```
plorenz, line 1973: Update description to "A service edge router policy update object"? Same for next one down?

specs/swagger.yml
```diff
@@ -3014,7 +3081,11 @@ responses:
   createResponse:
     description: The create request was successful and the resource has been added at the following location
     schema:
-      $ref: '#/definitions/create'
+      $ref: '#/definitions/createEnvelope'
+  sessionCreateResponse:
```
plorenz, line 3085: following

specs/swagger.yml
andrewpmartinez: Fixed

specs/swagger.yml
andrewpmartinez: Fixed

specs/swagger.yml
andrewpmartinez: fixed

PR #192, opened 2020-06-09
all conversion work

andrewpmartinez:
- remove old JSON schema, unsused code
- remove unused variables from base router functions
- rework all API entities
- rework all links
- fix lint issues
- adds error conversion logic
- update tests as needed
- fix up go test
- fix backwards compat w/ non json enroller


controller/embedded/api-schema/definitions.all.json
```diff
@@ -1,257 +0,0 @@
-{
```
andrewpmartinez, line 1: All JSON schema is no longer needed. Replaced w/ go-swagger generated validation.

controller/apierror/api.go
```diff
@@ -34,10 +38,71 @@ func (e ApiError) Error() string {
...
+			ret.Cause = &rest_model.APIErrorCause{
+				APIError: *causeApiError.ToRestModel(requestId),
+			}
+		} else if causeJsonSchemaError, ok := e.Cause.(*schema.ValidationErrors); ok {
+			//only possible from config type JSON schema validation
+			ret.Cause = &rest_model.APIErrorCause{
+				APIFieldError: rest_model.APIFieldError{
+					Field:  causeJsonSchemaError.Errors[0].Field,
+					Reason: causeJsonSchemaError.Errors[0].Error(),
+					Value:  fmt.Sprintf("%v", causeJsonSchemaError.Errors[0].Value),
```
dovholuknf, line 65: Errors[0] is guaranteed to exist? I found other places in here where [0] existed. i ask that you just reaffirm that they don't need boundary checking

tunnel/intercept/iputils.go
```diff
@@ -38,17 +38,17 @@ func getInterceptIP(hostname string, resolver dns.Resolver) (net.IP, error) {
 	addrs, err := net.LookupIP(hostname)
 	if err == nil {
 		if len(addrs) > 0 {
-			resolver.AddHostname(hostname, addrs[0].To4())
+			_ = resolver.AddHostname(hostname, addrs[0].To4())
```
dovholuknf, line 41: :( i'd kinda like to see at least these erorr logged but i get it - it's no worse than what we had

controller/env/appenv.go
```diff
@@ -183,8 +175,124 @@ type AppHandler func(ae *AppEnv, rc *response.RequestContext)
 
 type AppMiddleware func(*AppEnv, http.Handler) http.Handler
 
-func NewAppEnv(c *edgeconfig.Config) *AppEnv {
+type authorizer struct {
+}
+
+func (a authorizer) Authorize(request *http.Request, principle interface{}) error {
+	//principle is an API Session
```
dovholuknf, line 182: i actually can't discern if this should be principal or principle but i feel like it's probably principal?

controller/middleware/json.go
```diff
@@ -22,7 +22,7 @@ import (
 
 func SetResponseTypeToJson(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Set("Content-Type", "application/json")
+		w.Header().Set("content-type", "application/json")
```
dovholuknf, line 25: to be fair i always see this capitalized but headers are case insensitive so ¯\_(ツ)_/¯

controller/model/_delete.go
```diff
@@ -0,0 +1,15 @@
+package model
+
+import (
+	"github.com/openziti/foundation/transport/tcp"
+	"google.golang.org/appengine/socket"
+	"net/http"
+)
+
+func hi() {
```
dovholuknf, line 9: this might be my favorite function if this is intentional. does `hi` fulfill some interface? maybe a comment if this is useful in some way i'm not seeing?

controller/apierror/api.go
andrewpmartinez: It is, it is a composite error.

tunnel/intercept/iputils.go
andrewpmartinez: I was only attempting to satify the linter. I didn't know what the correct thing to do was so I left it as is.

controller/model/enrollment_mod_ca.go
```diff
@@ -181,8 +181,8 @@ func (module *EnrollModuleCa) Process(context EnrollmentContext) (*EnrollmentRes
 	return &EnrollmentResult{
 		Identity:      identity,
 		Authenticator: newAuthenticator,
-		Content:       []byte{},
-		ContentType:   "text/plain",
+		Content:       map[string]interface{}{},
+		TextContent:   []byte(""),
```
dovholuknf, line 185: I'm inferring that TextContent will implicitly set the Content-Type. if so - carry on - no response needed :) just pointing it out to make removing ContentType is intentional

controller/model/enrollment_mod_ott.go
```diff
@@ -99,11 +100,15 @@ func (module *EnrollModuleOtt) Process(ctx EnrollmentContext) (*EnrollmentResult
...
 
+	content := &rest_model.EnrollmentCerts{
+		Cert: string(certPem),
+	}
+
 	return &EnrollmentResult{
 		Identity:      identity,
 		Authenticator: newAuthenticator,
-		Content:       certPem,
-		ContentType:   "application/x-pem-file",
```
dovholuknf, line 106: hrmmm. I'm thinking the ContentType removal was purposeful now. I don't expect anything relied on the ContentType and it's perfectly fine to remove it

controller/middleware/json.go
andrewpmartinez: They are case insensitive by RFC. This was just a find/replace bystander.

controller/model/_delete.go
andrewpmartinez: Hahah I'll remove it.

controller/model/enrollment_mod_ca.go
andrewpmartinez: Text content can be many things, so yes it does have to set it.

This is a work-around for a larger issue that I didn't want to solve just now.

controller/schema/errors.go
```diff
@@ -27,7 +27,7 @@ type ValidationErrors struct {
 }
 
 func (e ValidationErrors) Error() string {
-	return fmt.Sprintf("schema validation failed")
+	return "schema validation failed"
```
dovholuknf, line 30: nothing of use in the `ValidationErrors` to return back to "whomever" gets this error? Pls confirm that this is "all the useful information"? :)

specs/swagger.yml
```diff
@@ -3083,6 +3083,7 @@ responses:
     schema:
       $ref: '#/definitions/createEnvelope'
   sessionCreateResponse:
+    description: The create request was successful and the resource has been added at the ollowing location.
```
dovholuknf, line 3086: is this really only an 'add' and not and 'add/delete'? pls look at this i feel like this is a mistake with `ollowing`

dovholuknf: the line below seems to be overriding this one?

controller/env/appenv.go
andrewpmartinez: I probable auto filled this out - so it is probably spelled that way in the code that defines the interface. Will check.

controller/internal/routes/ca_api_model.go
```diff
@@ -18,119 +18,103 @@ package routes
...
-		e.Links = &response.Links{
-			EntityNameSelf: e.GetSelfLink(),
-		}
+func MapCreateCaToModel(ca *rest_model.CaCreate) *model.Ca {
+	ret := &model.Ca{
+		BaseEntity: models.BaseEntity{
+			Tags: ca.Tags,
+		},
+		Name:                      stringz.OrEmpty(ca.Name),
+		Fingerprint:               "",
```
dovholuknf, line 69: i bet there is a reason why no fingerprint? maybe a comment would allow me to understand it. my instinct would be to fill it out

controller/internal/routes/edge_router_api_model.go
```diff
@@ -174,17 +150,176 @@ func MapEdgeRouterToApiList(ae *env.AppEnv, i *model.EdgeRouter) (*EdgeRouterApi
...
 		ret.EnrollmentToken = &enrollment.Token
 	}
 
-	ret.PopulateLinks()
-
 	return ret, nil
 }
+
+//
+//func NewEdgeRouterEntityRef(entity *model.EdgeRouter) *EntityApiRef {
```
dovholuknf, line 170: do you want to keep all this commented out code?

controller/internal/routes/enroll_router.go
```diff
@@ -105,51 +128,44 @@ func (ro *EnrollRouter) enrollHandler(ae *env.AppEnv, rc *response.RequestContex
...
-		err := json.Unmarshal(enrollmentResult.Content, &data)
-		if err != nil {
-			return err
+	//prefer json producer for non ott methods (backwards compat for ott)
+	explicitJsonAccept := false
+	if accept := rc.Request.Header.Values("accept"); len(accept) == 0 {
+		explicitJsonAccept = false //no headers specified
+	} else {
+		for _, val := range accept {
+			if strings.Split(val, ";")[0] == "application/json" {
```
dovholuknf, line 153: application/json is guaranteed to be at position 0? i don't know what `accept` will look like here so this is probably just fine again - just pointing it out to make sure

controller/internal/routes/identity_api_model.go
```diff
@@ -34,357 +36,246 @@ const (
...
-			Username: &username,
-		})
+	if identity.Enrollment != nil {
+		if identity.Enrollment.Ott {
+			enrollments = append(enrollments, &model.Enrollment{
+				BaseEntity: models.BaseEntity{},
+				Method:     persistence.MethodEnrollOtt,
+				Token:      uuid.New().String(),
+			})
+		} else if identity.Enrollment.Ottca != "" {
```
dovholuknf, line 80: i always always always trim the input "just in case". probably wouldn't hurt to trim `identity.Enrollment.Ottca` and `identity.Enrollment.Updb`?

controller/internal/routes/identity_api_model.go
```diff
@@ -34,357 +36,246 @@ const (
...
+			})
+		} else if identity.Enrollment.Updb != "" {
+			username := identity.Enrollment.Updb
+			enrollments = append(enrollments, &model.Enrollment{
+				BaseEntity: models.BaseEntity{},
+				Method:     persistence.MethodEnrollUpdb,
+				Token:      uuid.New().String(),
+				Username:   &username,
+			})
+		}
```
dovholuknf, line 96: should there be a final else and log to catch "just in case" situations should one ever arise?

controller/internal/routes/identity_router.go
```diff
@@ -46,187 +49,248 @@ func NewIdentityRouter() *IdentityRouter {
...
 		if err != nil {
 			return nil, err
 		}
 		result := make([]interface{}, 0)
-		for _, config := range configs {
-			result = append(result, IdentityServiceConfig{Service: config.Service, Config: config.Config})
+		for _, serviceConfig := range serviceConfigs {
+			service, err := ae.Handlers.EdgeService.Read(serviceConfig.Service)
+			if err != nil {
+				continue
```
dovholuknf, line 220: not even a debug log would be useful? if not - that's cool just askin'

controller/schema/errors.go
andrewpmartinez: Change was for the linter once again - said that the `Sprintf` wasn't needed. 

This specifically is a JSON marshaled object - so the logic to output useful information is in the `MarshalJSON` function which is a few lines above.

specs/swagger.yml
andrewpmartinez: Looks like an auto merge issue. Removed the extra line.

controller/internal/routes/edge_router_api_model.go
andrewpmartinez: Nope, was using them as reference. Forgot to remove this one.

controller/internal/routes/identity_api_model.go
andrewpmartinez: It isn't the APIs responsibility to trim input. The `Ottca` value is validated later by looking up the CA.

controller/internal/routes/identity_api_model.go
andrewpmartinez: No, the else is that the identity is being created w/o enrollments.

controller/internal/routes/identity_router.go
andrewpmartinez: A debug would be nice.

controller/apierror/messages.go
```diff
@@ -90,9 +90,11 @@ const (
 	InvalidEnrollMethodMessage string = "The supplied enrollment method is not valid"
 	InvalidEnrollMethodStatus  int    = http.StatusBadRequest
 
-	InvalidFilterCode    string = "INVALID_FILTER"
-	InvalidFilterMessage string = "The filter query supplied is invalid"
-	InvalidFilterStatus  int    = http.StatusBadRequest
+	StatusBadFilter                = 480
+	InvalidFilterCode       string = "INVALID_FILTER"
+	InvalidFilterMessage    string = "The filter query supplied is invalid"
+	httpStatusInvalidFilter        = StatusBadFilter
```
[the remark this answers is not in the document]
andrewpmartinez: Not anymore; got left behind. Removed extra const.

controller/env/broker.go
```diff
@@ -228,6 +228,7 @@ func (b *Broker) sendApiSessionCreates(apiSession *persistence.ApiSession) {
 	b.sendToAllEdgeRouters(channelMsg)
 }
```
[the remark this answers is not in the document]
andrewpmartinez: Updated todo

controller/env/appenv.go
andrewpmartinez: Wasn't - fixed in multiple locations besides this one.

controller/model/enrollment_mod_ca.go
andrewpmartinez: I removed my previous comment. I misread your comment.

The removal was on purpose - it was the start to a larger refactor that I put off.

The HTTP `content-type` is determined by whatever `Producer` is used to write the content. A `Producer` is suggested by the `go-swagger` generated code, but can it may be one of many valid choices.

At this point - it isn't up to the enrollment process to determine the content type. `TextContext` is a hamfisted workaround as I didn't want to refactor the enrollment modules at this time.

controller/model/enrollment_mod_ott.go
andrewpmartinez: Sure was see your other comment!

controller/internal/routes/ca_api_model.go
andrewpmartinez: It is handled at a lower level. 

The REST API `Map*` functions are largely meant to be dumb and defer to lower levels for logic and validation that can't be done on the REST API payload alone. Putting logic that logic here means any other presentation we use for the API (i.e. GRPC, channel2, etc.) would also have to have that logic.

controller/internal/routes/enroll_router.go
andrewpmartinez: This is another work-around for a larger issue - the same one I mentioned in another one of your comments dealing w/ `TextContent` in `EnrollmentResults`. 

This is essentially logic determines if we can use JSON as the response `content-type` at all and if we can prefer it. It is a concept that the `go-swagger` code doesn't attempt to solve for us - nor should it.

As for the format, it is specified by RFC 1049: `Content-Type:= type [";" ver-num [";" 1#resource-ref]] [comment]`

tests/edge_router_policy_test.go
```diff
@@ -50,57 +50,207 @@ func Test_EdgeRouterPolicy(t *testing.T) {
 	policy5 := ctx.AdminSession.requireNewEdgeRouterPolicyWithSemantic("AllOf", s("#"+edgeRouterRole1, "#"+edgeRouterRole2), s("#"+identityRole1, "#"+identityRole2))
 	policy6 := ctx.AdminSession.requireNewEdgeRouterPolicyWithSemantic("AnyOf", s("#"+edgeRouterRole1, "#"+edgeRouterRole2), s("#"+identityRole1, "#"+identityRole2))
```
plorenz, line 52: So tidy :heart_eyes:

controller/response/responder.go
```diff
@@ -17,179 +17,149 @@
...
+	"github.com/google/uuid"
 	"github.com/michaelquigley/pfxlog"
 	"github.com/openziti/edge/controller/apierror"
 	"github.com/openziti/edge/controller/schema"
+	"github.com/openziti/edge/internal/version"
+	"github.com/openziti/edge/rest_model"
 	"github.com/openziti/foundation/validation"
-	"github.com/sirupsen/logrus"
 	"net/http"
 )
```
plorenz, line 30: Todo still needed?

PR #200, opened 2020-06-15
Update rolling tests

plorenz:


plorenz:
Did some work trying to pull out tests into separate packages, for better visibility. Ended up not finishing that work because of additional complications with config file paths, etc, but figured it was ok to leave the refactoring making more things public.

PR #201, opened 2020-06-16
set base path to /edge/v1

andrewpmartinez:
- reroute API requests for the root to edge/v1 (i.e. / -> /edge/v1)
- fix pagination empty values being omitted

Author's Note: anything under `rest_server` `rest_client` are generated code. Reviewing swagger.yml will give you the source change.

### DIFF
--- a/controller/internal/routes/base_router.go
+++ b/controller/internal/routes/base_router.go
@@ -192,9 +192,9 @@ func ListWithEnvelopeFactory(rc *response.RequestContext, toEnvelope ApiListEnve
 
 	meta := &rest_model.Meta{
 		Pagination: &rest_model.Pagination{
-			Limit:      result.Limit,
-			Offset:     result.Offset,
-			TotalCount: result.Count,
+			Limit:      &result.Limit,
+			Offset:     &result.Offset,
+			TotalCount: &result.Count,
 		},
 		FilterableFields: result.FilterableFields,
 	}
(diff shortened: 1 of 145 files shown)


specs/swagger.yml
```diff
@@ -1,11 +1,11 @@
 ---
 swagger: '2.0'
 info:
```
plorenz, line 3: Would it be worth looking into updating this automatically as part of the build? Not in the context of this PR, but for future work..

controller/internal/routes/base_router.go
```diff
@@ -192,9 +192,9 @@ func ListWithEnvelopeFactory(rc *response.RequestContext, toEnvelope ApiListEnve
 
 	meta := &rest_model.Meta{
 		Pagination: &rest_model.Pagination{
```
plorenz, line 194: Did you have to switch these to pointers to get it to work, or was a stylistic preference?

specs/swagger.yml
andrewpmartinez: Right now the version tracks the Ziti release number but what are we going to do when we have v1, v2, ...vN REST API?

We could do something like `v1.0.19.0` `v2.0.19.0` or such.

controller/internal/routes/base_router.go
andrewpmartinez: I altered the swagger document to set the pagination fields as 'required' which `go-swagger` interprets as needing to be a pointer to check for nil vs zero value.

specs/swagger.yml
plorenz: hmm, that's interesting. Lots of options... We could keep them at the same version, but separate them by title. We could version them completely independently and only update the versions when the swagger file actually changes. That would probably be the most helpful to users, but the most work for us. We could also update a version in a comment or the title so that people could see what version of ziti it's associated with.

PR #202, opened 2020-06-16
Add.api.versions.to.versions

andrewpmartinez:
example:

```
{
    "data": {
        "apiVersions": {
            "edge": {
                "v1": {
                    "path": "/edge/v1"
                }
            }
        },
        "buildDate": "2020-06-11 16:03:13",
        "revision": "95e78d4bc64b",
        "runtimeVersion": "go1.14.3",
        "version": "v0.14.13"
    },
    "meta": {}
}
```

### DIFF
--- a/controller/internal/routes/version_router.go
+++ b/controller/internal/routes/version_router.go
@@ -19,6 +19,7 @@ package routes
 import (
 	"github.com/go-openapi/runtime/middleware"
 	"github.com/openziti/edge/build"
+	"github.com/openziti/edge/controller"
 	"github.com/openziti/edge/controller/env"
 	"github.com/openziti/edge/controller/internal/permissions"
 	"github.com/openziti/edge/controller/response"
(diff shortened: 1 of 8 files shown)


plorenz:
Is this available on an unversioned path? Even when we have /edge/v1, we probably want to reference this at just /version instead of /edge/v1/version

ekoby:
would it include unversioned path? as 
```
{
      "path": "",
      "version": "v0"
}
```

controller/internal/routes/version_router.go
```diff
@@ -53,14 +54,23 @@ func (ir *VersionRouter) Register(ae *env.AppEnv) {
...
 
 func (ir *VersionRouter) List(ae *env.AppEnv, rc *response.RequestContext) {
-
 	buildInfo := build.GetBuildInfo()
 	data := rest_model.Version{
 		BuildDate:      buildInfo.GetBuildDate(),
 		Revision:       buildInfo.GetRevision(),
 		RuntimeVersion: runtime.Version(),
 		Version:        buildInfo.GetVersion(),
+		APIVersions: []*rest_model.APIVersion{
```
ekoby, line 63: add unversioned paths/v0 for consistency?

andrewpmartinez: There is no v0. There is just v1 as we have it right now. It just happens to be temporarily available w/ and w/o path prefix.

The one w/o the path prefix is going to go away - it is deprecated. I don't see a reason to add it.

andrewpmartinez:
> Is this available on an unversioned path? Even when we have /edge/v1, we probably want to reference this at just /version instead of /edge/v1/version

Right now the API that is on the root is v1 and v1 is also on /edge/v1. So for now `GET /versions` is on the unversioned path.

When the fabric has its API and starts hosting the edge API, it will be responsible for the root `/version`

andrewpmartinez:
I updated the format so that we can have a map of different APIs to prepare for having the fabric in there too.

```{
    "data": {
        "apiVersions": {
            "edge": [
                {
                    "path": "/edge/v1",
                    "version": "v1"
                }
            ]
        },
        "buildDate": "2020-06-11 16:03:13",
        "revision": "95e78d4bc64b",
        "runtimeVersion": "go1.14.3",
        "version": "v0.14.13"
    },
    "meta": {}
}```

PR #203, opened 2020-06-16
Use.shortids

andrewpmartinez:
Added package `eid` in edge to mask what id generator is being used.

This is based off the `add.api.versions.toversions` branch so those changes will appear in this PR till that one merged.

shortid library:  https://github.com/teris-io/shortid

UUIDs are still used for CA verification/enrollment tokens/session tokens.


controller/internal/routes/ca_router.go
```diff
@@ -41,13 +41,11 @@ func init() {
 
 type CaRouter struct {
 	BasePath string
-	IdType   response.IdType
 }
 
 func NewCaRouter() *CaRouter {
 	return &CaRouter{
 		BasePath: "/" + EntityNameCa,
-		IdType:   response.IdTypeUuid,
```
andrewpmartinez, line 50: These were no longer being used. Removed from all REST API routers.

Issue #144, opened 2020-05-01
Unverified Edge Routers Cannot Be Used For Terminators


andrewpmartinez:
As edge router's do not create their fabric level router entry till, after enrollment, terminators cannot be assigned to them until then.

This is an issue for automation as creating a terminator may be done well before the enrollment process starts, let alone finishes.

PR #204, opened 2020-06-17
move edge routers under fabric routers. Fixes GH-144

plorenz:


controller/model/enrollment_mod_erott.go
```diff
@@ -143,7 +142,7 @@ func (module *EnrollModuleEr) Process(context EnrollmentContext) (*EnrollmentRes
 	}
 
 	cltFp := module.fingerprintGenerator.FromPem(cltPem)
-
+	cltFp = strings.Replace(strings.ToLower(cltFp), ":", "", -1)
```
andrewpmartinez, line 145: is there a reason this logic wasn't moved into fingerprintGenerator?

plorenz: fixed

Issue #206, opened 2020-06-18
Add snapshot db REST API


plorenz:
Add API to allow admin to trigger DB snapshot from REST

PR #207, opened 2020-06-18
Add snapshot db API. Fixes GH-206. Update openapi to 0.24

plorenz:


scripts/generate_rest.sh
```diff
@@ -2,7 +2,6 @@
 
 command -v swagger >/dev/null 2>&1 || {
   echo >&2 "Command 'swagger' not installed. See: https://github.com/go-swagger/go-swagger for installation"
-  exit 1
```
andrewpmartinez, line 5: ?

controller/internal/routes/database_router.go
```diff
@@ -0,0 +1,49 @@
...
+func (r *DatabaseRouter) Register(ae *env.AppEnv) {
+	ae.Api.DatabaseCreateDatabaseSnapshotHandler = database.CreateDatabaseSnapshotHandlerFunc(func(params database.CreateDatabaseSnapshotParams, _ interface{}) middleware.Responder {
+		return ae.IsAllowed(func(ae *env.AppEnv, rc *response.RequestContext) { r.CreateSnapshot(ae, rc) }, params.HTTPRequest, "", "", permissions.IsAdmin())
+	})
+}
+
+func (r *DatabaseRouter) CreateSnapshot(ae *env.AppEnv, rc *response.RequestContext) {
+	if err := ae.HostController.GetNetwork().SnapshotDatabase(); err != nil {
+		rc.RespondWithError(err)
+	}
```
andrewpmartinez, line 48: I would expect something like following based on the swagger responses defined:

```
func (r *DatabaseRouter) CreateSnapshot(ae *env.AppEnv, rc *response.RequestContext) {
	if err := ae.HostController.GetNetwork().SnapshotDatabase(); err != nil {
		rc.RespondWithError(err)
                return
	}
	rc.RespondWithEmptyOk()
}
```

specs/swagger.yml
```diff
@@ -2892,6 +2892,27 @@ paths:
...
+      security:
+        - ztSession: []
+      tags:
+        - Database
+      operationId: createDatabaseSnapshot
+      responses:
+        '200':
+          $ref: '#/responses/emptyResponse'
+        '400':
+          $ref: '#/responses/badRequestResponse'
```
andrewpmartinez, line 2912: If there are no inputs to this request, I don't believe it can return 400 bad request.

scripts/generate_rest.sh
plorenz: commented out b/c it wasn't detecting a swagger alias. I meant to put it back, fixing.

specs/swagger.yml
plorenz: Makes sense. It can fail if you invoke more than one per minute. That probably doesn't count as a bad request, though.

controller/internal/routes/database_router.go
plorenz: It seems to work without it. Maybe it defaults to 200 OK if nothing else is specified. I'll add it for clarity either way

andrewpmartinez: It might return 200 OK, but what is the payload?

specs/swagger.yml
andrewpmartinez: I suggest we detect that and throw a 409 conflict

controller/internal/routes/database_router.go
plorenz: Ok, makes sense. Appreciate the clarification. Code was previously updated to do ResponseWithEmptyOk

specs/swagger.yml
plorenz: I'll have to put in a fabric change to make the error detectable

PR #223, opened 2020-06-23
Fix listener close logic

plorenz:


gateway/xgress_edge/listener.go
```diff
@@ -232,17 +232,19 @@ func (proxy *ingressProxy) processBind(req *channel2.Message, ch channel2.Channe
 		if err := xgress.RemoveTerminator(proxy.listener.factory, terminatorId); err != nil {
```
plorenz, line 232: This file has the only actual fix. Everything else is test changes.

PR #234, opened 2020-07-01
fixes #233 return API errors as JSON if possible

### LINKED ISSUE #233: REST API Errors should be application/json if possible
The new REST API generated code performs default content negotiation and provides a default producer. During an API error, if `application/json` is acceptable to return, it should be. If not and a text producer is triggered, basic error information should be returned.

This issue arose during multiple consumptions of an enrollment token. The default text producer was being used to fulfill `application/x-pem-file` as a default content-type. During enrollment errors, this producer was still being used for API errors which would fail to produce any content.

In this situation the accept headers allows `application/json` but the default text producer was still being used. It should produce `application/json`. Additionally, if the accept header limited the response to `application/x-pem-file` the output should produce some error text.

andrewpmartinez:


tests/api_error_test.go
```diff
@@ -0,0 +1,103 @@
...
+			Post("enroll?token=" + madeUpToken)
+
+		ctx.Req.NoError(err)
+
+		contentTypeHeaders := resp.Header().Values("content-type")
+
+		ctx.Req.NotEmpty(contentTypeHeaders)
+		ctx.Req.Equal("application/x-pem-file", contentTypeHeaders[0])
+
+		ctx.Req.NotEmpty(resp.Body())
```
dovholuknf, line 101: i thought "basic error information should be returned."? should be able to do some kind of string search?

andrewpmartinez: This actually has to be worked on. Based on our call on Monday if `x-pem-file` is the content type then only empty or a PEM should be returned not textual error information

dovholuknf: i must not have been listening. i thought it was about returning json when pem was specified. if x-pem-file is supplied, imo, it's perfectly reasonable and very friendly to the consumer to return the error back in plain text. someone can write that into a file, try to use it and realize that it's not actually valid pem.

andrewpmartinez: It is technically wrong to misrepresent your content-type as @ekoby brought up.

If your client wants errors, it has to either provide an accept header that allows errors to be returned: `*/*` or `application/json, application/x-pem-file`.

Clients that don't still get an error status code.

dovholuknf: technically wrong - sure. but it's helpful to the consumer. with this implementation the consumer is just left in the dark and that's a much bigger issue than being technically correct. i will always argue on the side of being helpful. sometimes it's more important to be helpful than right.  ¯\_(ツ)_/¯

andrewpmartinez: Long term I want to remove the `x-pem-file` support and simply return JSON w/ PEM inside of it. That way the client doesn't have to worry about being wrong.

Issue #243, opened 2020-07-17
Edge Routers Returned On Session Create Is Limited To 10


andrewpmartinez:
*Experienced*

When creating a new session via the Edge API the returned list of Edge Routers is limited to 10. 

*Expected*

All should be returned

*Reproduction*

1. Create and run 11 Edge Routers associated to a service and identity
1. Enroll and identity and create a session for the service
1. Note that only 10 Edge Routers are returned

plorenz:
Should we add a parameter to /sessions to allow specifying how many edge routers to retrieve? Some SDK may want all, and some may want to limit it, in case there are hundreds. Or will the number of edge routers always be small enough that we will always want all?

ekoby:
> Should we add a parameter to /sessions to allow specifying how many edge routers to retrieve? Some SDK may want all, and some may want to limit it, in case there are hundreds. Or will the number of edge routers always be small enough that we will always want all?

yes

andrewpmartinez:
The fix in the above commit and merge should be fine as a short term solution. Should we get into a situation where we need to paginate/list/order the edge routers returned on session create we will cross that bridge when we get there.

PR #247, opened 2020-07-20
set edge router result list to all

andrewpmartinez:
- all is denoted by -1 which is actualy the max for int64


andrewpmartinez:
fixes #243

Issue #258, opened 2020-07-24
Integrate data integrity check/repair framework and add REST API


plorenz:
Integrate data integrity check/repair framework and add REST API

PR #259, opened 2020-07-24
Integrate data integrity check/repair framework and add REST API. Fix…

plorenz:
…es GH-258


specs/swagger.yml
```diff
@@ -2913,6 +2913,22 @@ paths:
           $ref: '#/responses/unauthorizedResponse'
         '429':
           $ref: '#/responses/rateLimitedResponse'
+  /database/checkDataIntegrity:
```
andrewpmartinez, line 2916: the API urls are all consistently kebab case not lower-camel case

`checkDataIntegrity` to `check-data-integrity`

specs/swagger.yml
```diff
@@ -2913,6 +2913,22 @@ paths:
...
+    post:
+      summary: Runs an data integrity scan on the datastore and returns any found issues
+      description: Runs an data integrity scan on the datastore and returns any found issues. Requires admin access.
+      security:
+        - ztSession: []
+      tags:
+        - Database
+      operationId: checkDataIntegrity
+      parameters:
+        - $ref: '#/parameters/fixErrors'
```
andrewpmartinez, line 2926: Seems odd to have a flag that fixes the error on the check. Seems odd to also have a `post` that is a read.

I'd suggest `POST /database/fix-data-integrity` and `GET /database/check-data-integrity`

plorenz:
Addressed review comments

PR #260, opened 2020-07-27
e2e support fixes #173

### LINKED ISSUE #173: Add e2e configuration to service model
- add an e2e encryption flag on service Ziti Edge service model
  - e2e flag should be immutable post-service creation  (initial concern as we don't have a great update system in place for service changes, avoid issues w/ clients and hosts agreeing on the settings)
- add support for a new xgress dialer, that is capable of dealing with e2e encryption (needs to be able to discover what services it can host and initialize necessary crypto)
- existing services will be upgraded with encryption on (requires the new xgress dialer)

andrewpmartinez:
- adds edge level edge_transport xgress
- adds service encyrptionRequired flag (rest API, protobuffs)
- adds edge router encryption checks
- adds encryption option to Ziti CLI for service create (default, true)
- updates Ziti CLI terminator creation to use edge_transport by default
- updates swagger and server/code gen

The code in this PR will not build initially. It requires changes in the fabric. See: https://github.com/openziti/fabric/pull/107


gateway/xgress_edge/dialer.go
```diff
@@ -101,5 +102,5 @@ func (dialer *dialer) Dial(destination string, sessionId *identity.TokenId, addr
 
 	start := edge.NewStateConnectedMsg(result.ConnId)
 	start.ReplyTo(reply)
```
plorenz, line 104: Can we return the PeerData here to make it work the same as the xgress_edge_transport impl?

gateway/xgress_edge/listener.go
```diff
@@ -150,13 +150,32 @@ func (proxy *ingressProxy) processConnect(req *channel2.Message, ch channel2.Cha
 	log.Debug("dialing fabric")
 	peerData := make(map[uint32][]byte)
 	peerData[edge.PublicKeyHeader] = req.Headers[edge.PublicKeyHeader]
+
+	if ns.Service.EncryptionRequired && req.Headers[edge.PublicKeyHeader] == nil {
```
plorenz, line 154: `conn.close(true, msg)` sends a state close reply, so you don't need both. I think we should be doing `conn.close(true, msg)` in the block above as well (if session create fails)

gateway/xgress_edge/dialer.go
andrewpmartinez: The public key that would need to be sent back that would make up the `peerData` is stored on the `terminator` that was added when the SDK client sent a bind request. So right now the public key is part of the static terminator peerData stored in the controller.

To do what you are asking would require changing storing the public key in memory on the router instead. 

Would that also trigger the removal of the peerData storage in the controller for the terimnator?

gateway/xgress_edge/listener.go
andrewpmartinez: There is a difference.

`conn.close` doesn't reply to the incoming `req` message. The `sendStateClosedReply` does by calling `replyFor()` on the state closed message.

Hence my confusion as in the original code both are used.

gateway/xgress_edge/dialer.go
andrewpmartinez: Will do this in a follow up.

andrewpmartinez: https://github.com/openziti/edge/issues/266

gateway/xgress_edge/listener.go
andrewpmartinez: Talked w/ Paul. Should `sendStateClosedReply` then `conn.close` w/o notify.

PR #270, opened 2020-08-04
Fix service policy types. Add debug util for debugging bbolt datastor…

plorenz:
…es. Fixes GH-269

### DIFF
--- a/util/debug-db/debug-db.go
+++ b/util/debug-db/debug-db.go
@@ -1,0 +1,110 @@
+package main
+
+import (
+	"fmt"
+	"github.com/openziti/edge/controller/model"
+	"github.com/openziti/edge/controller/persistence"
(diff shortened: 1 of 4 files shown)


util/debug-db/debug-db.go
```diff
@@ -0,0 +1,110 @@
+package main
```
plorenz, line 1: This is a command line util that you can use to add an admin/admin user to bbolt datastores you've gotten from users, so that you can debug it without having to ask for a valid username/password

andrewpmartinez: should it have a cmd folder?

plorenz: Maybe? I could also put it in ziti. Wasn't really sure where the best place was to put a debugging tool like this. Certainly don't want to ship it with ziti-all :)

PR #277, opened 2020-08-17
browzer-wss-underlay

rentallect:


gateway/xgress_edge_wss/dialer.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	if !result.Success {
+		return nil, fmt.Errorf("failed to establish connection with token %v. error: (%v)", token, result.Message)
+	}
+
+	conn := listenConn.newSink(result.NewConnId, dialer.options)
+
+	x := xgress.NewXgress(sessionId, address, conn, xgress.Terminator, &dialer.options.Options)
+	bindHandler.HandleXgressBind(sessionId, address, xgress.Terminator, x)
```
plorenz, line 99: You might be out of sync with the latest protocol change. I think there should be an x.Start() here

gateway/xgress_edge_wss/listener.go
```diff
@@ -0,0 +1,450 @@
+/*
```
plorenz, line 1: I feel like the vast majority of this code could be shared between implementations (I could be wrong, I haven't tried it). I'd be surprised if we couldn't pull at least some of this into a shared package, though

gateway/xgress_edge_wss/dialer.go
rentallect: The x.Start() is now in place.

gateway/xgress_edge_wss/listener.go
rentallect: I agree.  My initial goal was to get things flowing end-to-end (after being away from it for so long), using separate "wss" code, staying away from existing xgress_edge mechanisms.  We can now look into refactoring all xgress_edge_* code using shared code where possible.

gateway/xgress_edge_wss/dialer.go
```diff
@@ -0,0 +1,105 @@
+/*
```
andrewpmartinez, line 1: I don't think this xgress component needs to be able to dial. I think this whole thing can be blanked out.

gateway/xgress_edge_wss/listener.go
rentallect: deleted.  new wss transport is in place.

gateway/xgress_edge_wss/dialer.go
rentallect: deleted.  new wss transport is in place.

plorenz (approved):

Issue #281, opened 2020-08-19
Improve API Session and Session creation performance


plorenz:
* Remove identity foreign key index to api session to reduces writes during api session create
* Remove api session and service foreign key indexes to session to reduce writes during session create
* Ensure foreign key checks still happen (even though they are no longer indexed)
* Ensure we cascade delete when deleting identities, services and api sesions
* Make sdk and env info updates on identity batched when creating api sessions

PR #282, opened 2020-08-19
Improve session create performance. Closes GH-281

plorenz:


eid/id.go
```diff
@@ -1,10 +1,30 @@
...
 	"github.com/teris-io/shortid"
+	"math/rand"
+	"time"
 )
 
+var entropyC = make(chan string, 100)
+
+func init() {
+	go func() {
+		source := ulid.Monotonic(rand.New(rand.NewSource(time.Now().UnixNano())), 0)
```
andrewpmartinez, line 14: This library is interesting. It doesn't use lower case in the documentation example and states that its alphabet is `0123456789ABCDEFGHJKMNPQRSTVWXYZ` and says that `I, L, O and U are excluded` to avoid confusion. However, all uppercase letters are hard to read.

Also, these ids appear to be longer than what `shortid` provided. Is there a possibility of using a larger encoding alphabet (i.e. base64) to have shorter string representations?

PR #286, opened 2020-08-24
various fixes for API streaming and data net support

andrewpmartinez:
Requires a merge on fabric to compile: https://github.com/openziti/fabric/pull/120

- http requests should hard timeout w/ a reasonable response
- edge routers should close when an API session token is not provided or
  does not match
- stream both types of sessions on edge router connect
- stream sessions for time out processing/deletion
- add id to session messages (edge control pb update, breaks backwards
  compat)


controller/model/base_handler.go
```diff
@@ -186,6 +186,65 @@ func (handler *baseHandler) patchEntity(modelEntity boltEntitySource, checker bo
 	return handler.updateGeneral(modelEntity, checker, true)
 }
 
+func (handler *baseHandler) patchEntityBatch(modelEntity boltEntitySource, checker boltz.FieldChecker) error {
+	return handler.updateGeneralBatch(modelEntity, checker, true)
+}
+
+func (handler *baseHandler) updateGeneralBatch(modelEntity boltEntitySource, checker boltz.FieldChecker, patch bool) error {
```
andrewpmartinez, line 193: This is a stand-in copy of `updateGeneral()` with log output commented out.

`Batch`  writes to bolt can be rerun and thus produce extra log output. This implementation should evolve as we understand our needs for `Batch` writes.

plorenz: I was thinking about this again, and I feel like the log messages are ok, since they are reporting errors. I'm not clear on batch behavior, but I suspect that if your function returns an error, it should not be retried. Generally if it fails once, it will keep failing

gateway/xgress_edge/listener.go
```diff
@@ -158,7 +158,7 @@ func (proxy *ingressProxy) processConnect(req *channel2.Message, ch channel2.Cha
 		return
 	}
```
plorenz, line 160: I feel like we assume this is the token in some other places. Changing this may require a migration and updating some other bits of code. 

I'm not 100% sure of that, though. If you haven't vetted it yet, though, that would be something to look into

andrewpmartinez: What this affects is the `clientId` field in the fabric controller's in-memory session table that isn't persisted.

From the fabric's perspective, I saw it used to publish the fabric session create/delete events, but that was it.

controller/model/base_handler.go
andrewpmartinez: Uncommented, added "batch:" prefix to differentiate the log messages.

controller/model/base_handler.go
```diff
@@ -186,6 +186,62 @@ func (handler *baseHandler) patchEntity(modelEntity boltEntitySource, checker bo
...
+					}
+				} else {
+					pfxlog.Logger().Errorf("batch: entity of type %v is named, but store doesn't have name index", reflect.TypeOf(boltEntity))
+				}
+			}
+		}
+
+		if err := handler.GetStore().Update(ctx, boltEntity, checker); err != nil {
+			if patch {
+				pfxlog.Logger().Errorf("batch: entity of type %v is named, but store doesn't have name index", reflect.TypeOf(boltEntity))
```
dovholuknf, line 233: 233 and 236 could be put in front of the if block here

andrewpmartinez: Fixed

controller/env/broker.go
```diff
@@ -209,6 +286,10 @@ func (b *Broker) sendApiSessionCreates(apiSession *persistence.ApiSession) {
 		return
 	}
 
+	if len(fingerprints) == 0 {
+		return
```
dovholuknf, line 290: not even a debug here? feels like the sort of thing that might 'maybe' be useful some day? if not - i trust you :)

andrewpmartinez: Happens quite often for username/password logins. They don't have any certs. This line is specifically meant to weed them out.

Issue #291, opened 2020-09-01
Fix service policy denormalization migration


plorenz:
Policy type fix was done after migration, so migration skipped all service policies

PR #292, opened 2020-09-01
Fix service policy migration. Update for terminator precendence chang…

plorenz:
…es. Log panic stacks. Fixes GH-291


controller/timeout/timouts.go
```diff
@@ -63,15 +64,15 @@ func (h *timeoutHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	go func() {
 		defer func() {
 			if p := recover(); p != nil {
```
plorenz, line 66: @andrewpmartinez I added this because we used to get stackdumps on panics, and don't anymore. Let me know if you have any objections

andrewpmartinez: A perfectly logical change. I approve!

andrewpmartinez (approved):

Issue #226, opened 2020-06-24
Store identity and session IP addresses


andrewpmartinez:
As endpoint's interact with the controller and edge routers:

- log last known public IP address for the:
-- identity
-- API session
-- session

PR #295, opened 2020-09-09
adds public ip address for #226

andrewpmartinez:
- adds `http.Request.RemoteAddr` tracking to API sessions in a new field `ipAddress`
- `ipAddress` is IP only,  ipv4 or ipv6
- value can only be set via auth
- updates `swagger.yml` and generated code

### DIFF
--- a/controller/internal/routes/authenticate_router.go
+++ b/controller/internal/routes/authenticate_router.go
@@ -30,6 +30,7 @@ import (
 	"github.com/openziti/edge/rest_server/operations/authentication"
 	"github.com/openziti/foundation/metrics"
 	"github.com/openziti/foundation/util/stringz"
+	"net"
 	"net/http"
 	"time"
 )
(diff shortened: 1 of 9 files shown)


controller/internal/routes/authenticate_router.go
```diff
@@ -113,12 +114,17 @@ func (ro *AuthRouter) authHandler(ae *env.AppEnv, rc *response.RequestContext, p
 	if params.Body != nil {
 		configTypes = mapConfigTypeNamesToIds(ae, params.Body.ConfigTypes, identity.Id)
 	}
+	remoteIpStr := ""
+	if remoteIp, _, err := net.SplitHostPort(rc.Request.RemoteAddr); err == nil {
```
andrewpmartinez, line 118: `RemoteAddr` contains port info too

controller/persistence/api_session_store.go
```diff
@@ -49,13 +51,15 @@ func (entity *ApiSession) LoadValues(_ boltz.CrudStore, bucket *boltz.TypedBucke
 	entity.IdentityId = bucket.GetStringOrError(FieldApiSessionIdentity)
 	entity.Token = bucket.GetStringOrError(FieldApiSessionToken)
 	entity.ConfigTypes = bucket.GetStringList(FieldApiSessionConfigTypes)
+	entity.IPAddress = bucket.GetStringWithDefault(FieldApiSessionIPAddress, "")
```
andrewpmartinez, line 54: No migration needed. Post upgrade, any existing API Sessions will simply have blank IP addresses. New API Sessions will have the value populated.

andrewpmartinez:
Not sure if IPAddress should be in the standard output for `ziti edge list api-sessions`. Opinions accepted! Console output is pretty narrow and IPAddress may not be useful on every query.

gormami:
I would not put the IP in the standard output.  If there isn't a compelling reason, it shouldn't be everything, as it is hard to read.  But a switch on the command line to add it would be good, or a format type function where you could specify a formatting string of what you want to see would be great.

andrewpmartinez:
> I would not put the IP in the standard output. If there isn't a compelling reason, it shouldn't be everything, as it is hard to read. But a switch on the command line to add it would be good, or a format type function where you could specify a formatting string of what you want to see would be great.

There is a `-j` flag that allows the output to show everything in formatted JSON (to cover situations where there is way more info that is easy to format). Customizable formatting/return values I would like to see added, but is squarely in the nice-to-have category.

plorenz (approved):

PR #306, opened 2020-09-18
Support identity dial/bind

plorenz:


controller/env/broker.go
```diff
@@ -484,6 +484,7 @@ func (b *Broker) sendSessionCreates(session *persistence.Session) {
 		Service:          svc,
 		CertFingerprints: fps,
 		Type:             sessionType,
+		SessionToken:     session.ApiSession.Token,
```
andrewpmartinez, line 487: At first this looked like a bug because I saw an API Session Token being set as a Session Token.

The `Token` on 483 is a `SessionToken`  but the `Session` is kept off because of the context. I fee like this should be `APISessionToken`

andrewpmartinez: A second though - is there a reason the token is being sent instead of the id? Tokens are considered secrets and shouldn't be distributed unless required.

andrewpmartinez: A third thought - why does an edge router care about API sessions? At this point the concept of an API session has never made it to the router level.

andrewpmartinez: Looking over the code, it seems that the API Session token is being used in a few places and being sent to the routers. This should probably be `id` only.

andrewpmartinez: To be clear, a compromised Session Token doesn't get someone very far. To use it for anything they would also have to have the client key pair. An API Session Token that is compromised can give someone elevated access to the management API.

andrewpmartinez (changes requested):
See comments about api session tokens. If we want to punt this to a separate issue to move forward - that is fine. It isn't a security hole per-se but it does make it significantly easier for an attacker on a compromised machine.

controller/env/broker.go
plorenz: This field isn't used, as far as I can tell. I started to use it, noticed it wasn't being populated, then later removed the use. I think we can just drop the field. 

We are also using token on the ApiSession, and we're using it to validate the initial connection to the gateway from the sdk client. I suspect switching that to Id wouldn't make any sense, since we'd just be making the id the effective token.

andrewpmartinez: Ah yes, incoming API session validation so that connections can't be zombies. Now I remember.

andrewpmartinez (approved):

PR #310, opened 2020-09-22
Fix.305.cert.auth.ca.validation

andrewpmartinez:
- cert auth module caches static and dynamic (3rd party) CA chains
- all cert auths validated against active chains
- deleted/disabled 3rd party CAs no longer validate

### DIFF
--- a/controller/model/authenticator_mod_cert.go
+++ b/controller/model/authenticator_mod_cert.go
@@ -24,6 +24,8 @@ import (
 	"github.com/openziti/edge/controller/apierror"
 	"github.com/openziti/edge/controller/persistence"
 	"github.com/openziti/edge/internal/cert"
+	nfpem "github.com/openziti/foundation/util/pem"
+	cmap "github.com/orcaman/concurrent-map"
 	"net/http"
 )
 
@@ -36,14 +38,20 @@ type AuthModuleCert struct {
 	env                  Env
 	method               string
 	fingerprintGenerator cert.FingerprintGenerator
+	caChain              []byte
+	staticCaCerts        []*x509.Certificate
+	dynamicCaCache       cmap.ConcurrentMap //map[string][]*x509.Certificate
 }
 
-func NewAuthModuleCert(env Env) *AuthModuleCert {
+func NewAuthModuleCert(env Env, caChain []byte) *AuthModuleCert {
 	handler := &AuthModuleCert{
 		env:                  env,
 		method:               persistence.MethodAuthenticatorCert,
 		fingerprintGenerator: cert.NewFingerprintGenerator(),
+		staticCaCerts:        nfpem.PemToX509(string(caChain)),
+		dynamicCaCache:       cmap.New(),
 	}
+
 	return handler
 }
 
@@ -58,7 +66,7 @@ func (module *AuthModuleCert) Process(context AuthContext) (string, error) {
 		return "", err
 	}
 
-	for fingerprint := range fingerprints {
+	for fingerprint, cert := range fingerprints {
 		authenticator, err := module.env.GetHandlers().Authenticator.ReadByFingerprint(fingerprint)
 
 		if err != nil {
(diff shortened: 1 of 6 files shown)


controller/model/authenticator_mod_cert.go
```diff
@@ -81,13 +89,55 @@ func (module *AuthModuleCert) Process(context AuthContext) (string, error) {
...
+			opts := x509.VerifyOptions{
+				Roots:         module.getRootPool(),
+				Intermediates: x509.NewCertPool(),
+				KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageAny},
+			}
+
+			if _, err := cert.Verify(opts); err == nil {
+				return authenticator.IdentityId, nil
+			} else {
+				println(err)
```
dovholuknf, line 101: shouldn't this be a log msg and not a println?

controller/model/authenticator_mod_cert.go
```diff
@@ -81,13 +89,55 @@ func (module *AuthModuleCert) Process(context AuthContext) (string, error) {
...
 }
 
+func (module *AuthModuleCert) getRootPool() *x509.CertPool {
+	roots := x509.NewCertPool()
+
+	for _, caCert := range module.staticCaCerts {
+		roots.AddCert(caCert)
+	}
+
+	cas, err := module.env.GetHandlers().Ca.Query("isAuthEnabled = true and isVerified = true limit 500")
```
dovholuknf, line 116: i can't imagine a world where there's 500 ca's in one instance of ziti but... dumb things happen maybe? is there any harm in going moar bigger? why stop at 500 when it seems (to me from the code) that this is "just give me all of them"

controller/model/authenticator_mod_cert.go
```diff
@@ -81,13 +89,55 @@ func (module *AuthModuleCert) Process(context AuthContext) (string, error) {
...
+
+	if err != nil {
+		return nil
```
dovholuknf, line 119: shouldn't we log this err "just in case"?

tests/ca_test.go
```diff
@@ -215,4 +215,70 @@ func Test_CA(t *testing.T) {
...
+
+		t.Run("deleting a CA no longer allows authentication", func(t *testing.T) {
+			ctx.testContextChanged(t)
+			ctx.AdminSession.requireDeleteEntity(ca)
+
+			enrolledSession, err := clientAuthenticator.Authenticate(ctx)
+
+			ctx.Req.Error(err)
+			ctx.Req.Empty(enrolledSession)
+		})
```
dovholuknf, line 282: does it make sense to add a test here for "client has a session, remove the CA, session is removed/disconnected"? or is that somehow covered in these tests already?

dovholuknf (approved):

controller/model/authenticator_mod_cert.go
andrewpmartinez: Leftover from debugging, removed.

controller/model/authenticator_mod_cert.go
smilindave26: data net enrolls every customer's CA as 3rd party ca and auto-enrolls the probes

andrewpmartinez: Should be a stream of CAs, fixing.

controller/model/authenticator_mod_cert.go
andrewpmartinez: changed when moving to a streaming call

tests/ca_test.go
andrewpmartinez: I looked through this and I'm adding another ticket to address this.

The system doesn't do that. Pretty sure it never did. Also, there isn't a simple lookup to determine which API sessions to delete, but it can be added. I also was thinking if we disable auth on a CA does that also terminate existing API sessions and sessions?

dovholuknf: seems like it should to me. might be a footgun but if it's easy enough to do then I think it makes sense

andrewpmartinez: https://github.com/openziti/edge/issues/311

PR #327, opened 2020-10-05
fixes #326 nested tx for session timeouts

### LINKED ISSUE #326: Session timeout can deadlock
Session timeout can deadlock with a read/delete session nesting.

andrewpmartinez:


tunnel/intercept/protocols/tcp/conn.go
```diff
@@ -59,7 +59,7 @@ type ClientConn struct {
 	rcvAckNum   uint32
 	sndUna      uint32 // oldest unacknowledged sequence number
 	sndAckNum   uint32 // bytes received. written in tcpRecv. read in tcpSend
-	seqCond     sync.Cond
+	seqCond     *sync.Cond
```
andrewpmartinez, line 62: `go vet` was complaining that mutex's were being copied by value and not by reference -.-

controller/internal/policy/session.go
```diff
@@ -49,10 +54,28 @@ func NewSessionEnforcer(appEnv *env.AppEnv, frequency time.Duration, sessionTime
...
+		if err != nil {
+			pfxlog.Logger().Errorf("encountered error querying for API sessions to remove: %v", err)
+			break
+		}
+
+		if len(ids) == 0 {
+			break
+		}
+
+		for _, id := range ids {
```
plorenz, line 75: This is still a nested tx. We probably need to expose the tx in the func passed to streamIds.

plorenz (changes requested):
Have delete use same tx as StreamIds or don't use streamIds, use QueryIds

controller/internal/policy/session.go
andrewpmartinez: How is it nested? the `StreamIds` bundles the ids up, view tx closed, then delete in separate txs.

plorenz: Ah, I missed that the delete wasn't in the stream ids func anymore. I still think it should be using QueryIds, since you've essentially reimplemented QueryIds using StreamIds :) But problem is fixed.

plorenz (approved):

PR #346, opened 2020-10-13
Narrow json types for list items

scareything:
Attempts to create a service configuration with a list of numbers would fail with `unsupported type json.Number in map`. this PR narrows the json types for lists.

### DIFF
--- a/controller/internal/routes/config_api_model.go
+++ b/controller/internal/routes/config_api_model.go
@@ -44,7 +44,7 @@ func MapCreateConfigToModel(config *rest_model.ConfigCreate) *model.Config {
 	dataMap := config.Data.(map[string]interface{})
 	ret.Data = dataMap
 
-	narrowJsonTypes(ret.Data)
+	narrowJsonTypesMap(ret.Data)
 
 	return ret
 }
@@ -62,7 +62,7 @@ func MapUpdateConfigToModel(id string, config *rest_model.ConfigUpdate) *model.C
 		ret.Data = dataMap
 	}
 
-	narrowJsonTypes(ret.Data)
+	narrowJsonTypesMap(ret.Data)
 
 	return ret
 }
@@ -80,7 +80,7 @@ func MapPatchConfigToModel(id string, config *rest_model.ConfigPatch) *model.Con
 		ret.Data = dataMap
 	}
 
-	narrowJsonTypes(ret.Data)
+	narrowJsonTypesMap(ret.Data)
 
 	return ret
 }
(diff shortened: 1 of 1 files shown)


controller/internal/routes/config_api_model.go
```diff
@@ -125,27 +125,47 @@ func MapConfigToRestModel(ae *env.AppEnv, config *model.Config) (*rest_model.Con
...
 		}
+	}
+}
+
+func narrowJsonTypesList(l []interface{}) {
+	for i, v := range l {
+		narrowJsonType(&v)
 
 		switch val := v.(type) {
-		case float64:
```
scareything, line 142: I don't really understand why this was being done here instead of when parsedNumber.Float64() succeeded. @andrewpmartinez

andrewpmartinez: Because the above section is gated by a type assertion check and it may or may not be hit.

IIRC this was to deal with how `go-swagger` was dealing with maps that contained values where it explicitly didn't know the type to give to a number because the config type definition is defined as "a map of things to values" which denotes no specific types for number-like values.

`go-swagger` instead could either guess or give us an interface object w/ methods to figure it out on our own. The below interface `ParsedNumber` is a sub-interface of that interface and used to detect that scenario.

andrewpmartinez: Per my comment above I don't know if anything changed that makes it so we don't need to look for `ParsedNumber` anymore. 

This new logic assumes we have `int64`/`float64` every time and not `ParsedNumber`.

controller/internal/routes/config_api_model.go
```diff
@@ -125,27 +125,47 @@ func MapConfigToRestModel(ae *env.AppEnv, config *model.Config) (*rest_model.Con
 	return ret, nil
 }
```
plorenz, line 127: This is a style thing, so feel free to ignore, but I would prefer `func narrowJsonType(v interface{}) interface{}`

controller/internal/routes/config_api_model.go
```diff
@@ -125,27 +125,47 @@ func MapConfigToRestModel(ae *env.AppEnv, config *model.Config) (*rest_model.Con
...
-		if parsedNumber, ok := v.(ParsedNumber); ok {
-			//floats don't parse as int, try int fist, then float, else give up
-			if intVal, err := parsedNumber.Int64(); err == nil {
-				v = intVal
-				m[k] = intVal
-			} else if floatVal, err := parsedNumber.Float64(); err == nil {
-				v = floatVal
-				m[k] = floatVal
+func narrowJsonType(v *interface{}) {
+	if parsedNumber, ok := (*v).(ParsedNumber); ok {
```
andrewpmartinez: I would rename this function to `resolveParsedNumber` or something to the like.

andrewpmartinez (approved):

plorenz (approved):

PR #355, opened 2020-10-15
fix #349: avoid putting empty key into peerData

### LINKED ISSUE #349: Using 'edge_transport' binding on non-encrypted service leads to Dial failure
If you create a service that has `encryptionRequired:false`, and then create a terminator and Include `"binding" : "edge_transport"` in the POST payload given to the `/terminators` endpoint, the terminator is created, but it will lead to a down-stream Dial failure when the SDK client attempts to connect to the service.

Here is an example error seen in smoketest:
```
{
    "file": "/home/travis/gopath/pkg/mod/github.com/openziti/edge@v0.16.35/tunnel/tunnel.go:31",
    "func": "github.com/openziti/edge/tunnel.DialAndRun",
    "level": "error",
    "msg": "zt.Dial(ziti_smoketest_files) failed: unable to dial service 'ziti_smoketest_files' (attempt to use closed connection: error creating route for [s/xyPp] (error preparing e2e crypto: failed key exchange: bad point length: 0, expected 32))",
    "time": "2020-10-14T17:09:20Z"
}
```

ekoby:


dovholuknf (approved):

router/xgress_edge/listener.go
```diff
@@ -153,7 +153,10 @@ func (proxy *ingressProxy) processConnect(req *channel2.Message, ch channel2.Cha
 	// fabric connect
 	log.Debug("dialing fabric")
 	peerData := make(map[uint32][]byte)
-	peerData[edge.PublicKeyHeader] = req.Headers[edge.PublicKeyHeader]
+
+	if pk, found := req.Headers[edge.PublicKeyHeader]; found {
```
andrewpmartinez, line 157: I found this too, but wasn't able to confirm the fix. I couldn't reproduce the situation.

I also found that sometimes the header was being sent in w/ a length of zero.

andrewpmartinez: My current version was:

```
	if pk, found := req.Headers[edge.PublicKeyHeader]; found && len(pk) > 0{
		peerData[edge.PublicKeyHeader] = pk
	}
```

andrewpmartinez: But it didn't feel right. The header shouldn't be sent if it is going to be empty.

plorenz (approved):

PR #384, opened 2020-10-28
posture check implementation

andrewpmartinez:


controller/model/posture_check_model_process.go
```diff
@@ -23,12 +23,41 @@ import (
...
+func (p *PostureCheckProcess) Evaluate(pd *PostureData) bool {
+	for _, process := range pd.Processes {
+		if process.PostureCheckId == p.PostureCheckId {
+			if process.TimedOut {
+				return false
+			}
+
+			if p.Fingerprint != "" {
+				if p.Fingerprint != process.SignerFingerprint {
+					return false
```
dovholuknf, line 42: shouldn't this just 'continue' and not return? it's in a for loop - don't you need to check them all before returning a response?

controller/model/posture_check_model_process.go
```diff
@@ -23,12 +23,41 @@ import (
...
+			}
+
+			if len(p.Hashes) > 0 {
+				for _, hash := range p.Hashes {
+					if hash == process.BinaryHash {
+						return true
+					}
+				}
+			} else {
+				return true
```
dovholuknf, line 53: if the PostureCheckProcess has no hashes (p.Hashes > 0) and that's a success - can you move the if len(p.Hashes) <= 0 check outside the for loop and speed up the checking process a smidgeon and make this code here a smidge tighter?

controller/model/posture_check_model_windows_domain.go
```diff
@@ -26,6 +26,20 @@ type PostureCheckWindowsDomains struct {
 	Domains []string
 }
 
+func (p *PostureCheckWindowsDomains) Evaluate(pd *PostureData) bool {
+	if pd.Domain.TimedOut {
+		return false
```
dovholuknf, line 31: on all these TimedOut checks, is it worthwhile to log/know when those types of timeouts occur? not sure if it is exceptional, expected, or if it could be a flood of logs so i can understand if it's not logged on purpose but seems like that sort of information will be hard to debug (maybe)

controller/model/posture_check_model_os.go
```diff
@@ -26,6 +26,28 @@ type PostureCheckOperatingSystem struct {
 	OperatingSystems []OperatingSystem
 }
 
+func (p *PostureCheckOperatingSystem) Evaluate(pd *PostureData) bool {
+	if pd.Os.TimedOut {
+		return false
+	}
+
+	validOses := map[string]map[string]struct{}{}
```
dovholuknf, line 34: can validOses be done in an init or does it need to be done every call to Evaluate? It doesn't seem to use the posture data supplied which makes me think it might be eligible to be done via init instead?

controller/model/posture_check_model_mac.go
```diff
@@ -26,6 +26,25 @@ type PostureCheckMacAddresses struct {
 	MacAddresses []string
 }
 
+func (p *PostureCheckMacAddresses) Evaluate(pd *PostureData) bool {
+	if pd.Mac.TimedOut {
+		return false
+	}
+
+	validAddresses := map[string]struct{}{}
```
dovholuknf, line 34: same comment wrt creating the validAddresses map - can it be done outside the Evaluate?

controller/internal/routes/posture_check_router.go
```diff
@@ -66,7 +69,49 @@ func (r *PostureCheckRouter) Register(ae *env.AppEnv) {
...
+
+			apiEntities, err = modelToApi(ae, rc, MapPostureCheckToRestEntity, result.GetEntities())
+			if err != nil {
+				return nil, err
+			}
+			qmd = &result.QueryMetaData
+		} else {
+			result, err := ae.Handlers.PostureCheck.QueryPostureChecks(query)
+			if err != nil {
+				pfxlog.Logger().Errorf("error executing list query: %+v", err)
```
dovholuknf, line 104: why is this one error logged when all the others aren't - does this need to be logged here vs the others?

andrewpmartinez: debug I left in.

controller/model/posture_check_model_process.go
andrewpmartinez: If there is a signer fingerprint defined, and it is missing, the posture fails. There is no point in continuing.

controller/model/posture_check_model_process.go
andrewpmartinez: I don't think that will improve anything. Readability maybe but that wasn't brought up.

controller/model/posture_check_model_windows_domain.go
andrewpmartinez: Timeouts are expected to happen if the client stops sending data. The frequency could be large or very little.

If you want to know if you have timed out data you can review the posture data on any identity and it will clearly show the data timed out.

dovholuknf (approved):

controller/model/posture_check_model_mac.go
andrewpmartinez: The PostureCheckMacAddresses `p` is coming from bbolt and I don't see a great place to put this logic such that it will improve anything. 

Something I wanted to add, but didn't in this interaction, was an evaluation cache that would be smart enough to know if the data actually updated or if it just refreshed. That would actually be a bigger improvement.

controller/model/posture_check_model_process.go
andrewpmartinez: I think I see your confusion.

PostureData `pd` can have multiple processes reported from the client. In this evaluation, we are looking for the single reported process in `pd.Processes` by looking at the `PostureCheckId`. The loop should only hit that `PostureCheckId` comparison once and when it does it is ride or die, hence all the returns.

controller/model/posture_check_model_os.go
andrewpmartinez: See the other comment! I replied in a different place.

PR #388, opened 2020-10-30
fixes for mop/sdks

andrewpmartinez:
- case insensitive checks for domain/os
- no minimum mac addresses on posture responses


controller/model/posture_check_model_windows_domain.go
```diff
@@ -32,7 +33,7 @@ func (p *PostureCheckWindowsDomains) Evaluate(pd *PostureData) bool {
 	}
 
 	for _, domain := range p.Domains {
```
plorenz, line 35: I like strings.EqualFold, but it's a matter of taste.

plorenz (approved):

PR #389, opened 2020-11-01
guard double close of channel

ekoby:


router/xgress_edge_transport/connection.go
```diff
@@ -185,6 +185,7 @@ func (c *edgeTransportXgressConn) WritePayload(p []byte, headers map[uint8][]byt
 
 	if err != nil {
 		pfxlog.ContextLogger(c.LogContext()).WithError(err).Error("write failed, closing connection")
+		c.finSent = true
 		close(c.writeDone)
```
andrewpmartinez, line 189: Do we need

`		if !c.finSent {
			c.finSent = true
			close(c.writeDone)
		}
`

andrewpmartinez: I ask because the panic seemed to be on line 188.

ekoby: c.finSent is checked first thing in this function

andrewpmartinez (approved):

PR #399, opened 2020-11-11
process checks are case insensitive and remove isSigned

andrewpmartinez:
- isSigned is removed as it is not used
- process binary and signer hashes are case insensitive

### DIFF
--- a/controller/model/posture_check_model_process.go
+++ b/controller/model/posture_check_model_process.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"github.com/openziti/edge/controller/persistence"
 	"go.etcd.io/bbolt"
+	"strings"
 )
 
 type PostureCheckProcess struct {
@@ -42,14 +43,14 @@ func (p *PostureCheckProcess) Evaluate(pd *PostureData) bool {
 			}
 
 			if p.Fingerprint != "" {
-				if p.Fingerprint != process.SignerFingerprint {
+				if !strings.EqualFold(p.Fingerprint, process.SignerFingerprint) {
 					return false
 				}
 			}
 
 			if len(p.Hashes) > 0 {
 				for _, hash := range p.Hashes {
-					if hash == process.BinaryHash {
+					if strings.EqualFold(hash, process.BinaryHash) {
 						return true
 					}
 				}
--- a/controller/model/posture_check_model_process_test.go
+++ b/controller/model/posture_check_model_process_test.go
@@ -185,7 +185,7 @@ func TestPostureCheckModelProcess_Evaluate(t *testing.T) {
 		processCheck, postureData := newMatchingProcessCheckAndData()
 		postureData.Processes[0].IsRunning = false
 		postureData.Processes[0].BinaryHash = "does not match"
-		postureData.Processes[0].BinaryHash = "does not match"
+		postureData.Processes[0].SignerFingerprint = "does not match"
 
 		result := processCheck.Evaluate(postureData)
 
--- a/controller/model/posture_check_model_process_test.go
+++ b/controller/model/posture_check_model_process_test.go
@@ -1,0 +1,195 @@
+/*
+	Copyright NetFoundry, Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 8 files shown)


specs/swagger.yml
```diff
@@ -6631,8 +6631,6 @@ definitions:
         properties:
           isRunning:
             type: boolean
-          isSigned:
```
dovholuknf, line 6634: any chance this will cause mop a regression of any flavor? i doubt it but figured i'd ask. almost seems 'breaking ish'

controller/persistence/posture_check_process.go
```diff
@@ -51,6 +52,13 @@ func (entity *PostureCheckProcess) LoadValues(_ boltz.CrudStore, bucket *boltz.T
 }
 
 func (entity *PostureCheckProcess) SetValues(ctx *boltz.PersistContext, bucket *boltz.TypedBucket) {
+
+	entity.Fingerprint = strings.ToLower(entity.Fingerprint)
```
dovholuknf, line 56: are fingerprints ever allowed to have whitespace or other delimiters in them? should that be normalized out if yes?

controller/model/posture_check_model_process.go
```diff
@@ -42,14 +43,14 @@ func (p *PostureCheckProcess) Evaluate(pd *PostureData) bool {
 			}
 
 			if p.Fingerprint != "" {
-				if p.Fingerprint != process.SignerFingerprint {
+				if !strings.EqualFold(p.Fingerprint, process.SignerFingerprint) {
```
dovholuknf, line 46: same general questions around 'normalizing' fingerprints and hashes. can they have whitespace, can they have separators? if yes should that all be normalized? (maybe it is already)

controller/model/posture_check_model_process_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+		req := require.New(t)
+		req.False(result)
+	})
+
+	t.Run("returns false not running, invalid hash, invalid signer", func(t *testing.T) {
+		processCheck, postureData := newMatchingProcessCheckAndData()
+		postureData.Processes[0].IsRunning = false
+		postureData.Processes[0].BinaryHash = "does not match"
+		postureData.Processes[0].BinaryHash = "does not match"
```
dovholuknf, line 188: duplicated line that doesn't have any effect :)

controller/model/posture_check_model_process_test.go
```diff
@@ -0,0 +1,195 @@
...
+		result := processCheck.Evaluate(postureData)
+		req := require.New(t)
+
+		req.True(result)
+	})
+
+	t.Run("returns true for valid id and running but check has null hashes and no signer", func(t *testing.T) {
+		processCheck, postureData := newMatchingProcessCheckAndData()
+		processCheck.Hashes = nil
+		processCheck.Fingerprint = ""
```
dovholuknf, line 115: what if the fingerprint gets leading/trailing whitespace? might be worth testing?

dovholuknf (approved):

specs/swagger.yml
andrewpmartinez: Nope.

controller/persistence/posture_check_process.go
andrewpmartinez: Valid characters are `a-f, 0-9`. Could be normalized. I'll add a ticket for this because the work around is to do the right thing.

andrewpmartinez: https://github.com/openziti/edge/issues/400

controller/model/posture_check_model_process_test.go
andrewpmartinez: Fixed.

controller/model/posture_check_model_process_test.go
andrewpmartinez: I'll address in #400.

controller/model/posture_check_model_process.go
andrewpmartinez: Will address in #400

PR #402, opened 2020-11-12
Enhance OS version posture checks

### LINKED ISSUE #401: add flexible major minor patch version posture checks of os
Allow version checks for OS posture checks to work off of a range of semver based ranges on x.y.z versions.

andrewpmartinez:
Valid x.y.z versions of OS posture checks can now be specified by
posture check ranges.

### DIFF
--- a/controller/model/posture_check_model_os_test.go
+++ b/controller/model/posture_check_model_os_test.go
@@ -1,0 +1,304 @@
+/*
+	Copyright NetFoundry, Inc.
+
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
+
+	https://www.apache.org/licenses/LICENSE-2.0
+
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package model
+
+import (
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+)
+
+func TestPostureCheckModelOs_Evaluate(t *testing.T) {
+
+	t.Run("returns true for exactly matching valid os type and version", func(t *testing.T) {
+		osCheck, postureData := newMatchingOsCheckAndData()
+
+		result := osCheck.Evaluate(postureData)
+
+		req := require.New(t)
+		req.True(result)
+	})
+
+	t.Run("returns true for valid os type and higher than required major version", func(t *testing.T) {
+		osCheck, postureData := newMatchingOsCheckAndData()
+		osCheck.OperatingSystems[0].OsVersions[0] = ">=1.0.0"
+
+		result := osCheck.Evaluate(postureData)
+
+		req := require.New(t)
+		req.True(result)
+	})
+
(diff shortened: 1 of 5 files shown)


controller/model/posture_check_model_os_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+		result := osCheck.Evaluate(postureData)
+
+		req := require.New(t)
+		req.False(result)
+	})
+
+	t.Run("returns false for posture data with valid os and partial os major and minor version match with dangling period", func(t *testing.T) {
+		osCheck, postureData := newMatchingOsCheckAndData()
+		postureData.Os.Version = "10.5."
```
dovholuknf, line 204: how about a `10..5` test too? and `.5.5` and `0.2.0` or `0.0.0` maybe?

controller/model/posture_check_model_os_test.go
```diff
@@ -0,0 +1,304 @@
...
+		osCheck, postureData := newMatchingOsCheckAndData()
+		postureData.Os.Version = "10.5"
+
+		result := osCheck.Evaluate(postureData)
+
+		req := require.New(t)
+		req.False(result)
+	})
+
+	t.Run("returns false for posture data with valid os and partial os major version match with dangling period", func(t *testing.T) {
```
dovholuknf, line 192: this doesn't have a 'dangling' period does it?

controller/model/posture_check_model_os_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+		result := osCheck.Evaluate(postureData)
+
+		req := require.New(t)
+		req.False(result)
+	})
+
+	t.Run("returns false for posture data with valid os and partial os major and minor version match", func(t *testing.T) {
+		osCheck, postureData := newMatchingOsCheckAndData()
+		postureData.Os.Version = "10.5"
```
dovholuknf, line 184: mac os might be 14.1 and 14.2 ? is the minor really necessary? ==> 14.1.0 ,14.2.0 ?

controller/model/posture_check_model_os_test.go
andrewpmartinez: At that point, we are testing the edge cases of the semver library I am using. I'd rather leave that testing in their court. They will all fail.

controller/model/posture_check_model_os_test.go
andrewpmartinez: Sure doesn't

controller/model/posture_check_model_os_test.go
andrewpmartinez: I looked up their versions and they have a patch number.

andrewpmartinez: https://stackoverflow.com/questions/157759/how-can-i-determine-the-running-mac-os-x-version-programmatically

smilindave26: yes - mac has a patch number.  I'm not currently sending it up since MOP isn't configured for it, and this (version x.y.z of higher") wasn't yet supported.

iOS doesn't supply a patch number in the version.

dovholuknf (approved):

controller/model/posture_check_model_os_test.go
andrewpmartinez: We should send in 0 where it doesn't exist. Otherwise, I have to introduce special cases.

smilindave26: okay, no problem

andrewpmartinez: 😎

PR #407, opened 2020-11-12
fixes #406 adds WindowsServer in addition to Windows

### LINKED ISSUE #406: Enhance OS posture checks to support windows server vs windows desktop

andrewpmartinez:
fixes #406


dovholuknf (approved):

specs/swagger.yml
```diff
@@ -3890,6 +3890,7 @@ definitions:
     type: string
     enum:
       - Windows
+      - WindowsServer
```
andrewpmartinez, line 3893: This line allows WindowsServer as an OS type in the API. The rest of the back end assumes the validation has properly filtered values and doesn't inspect them past string comparisons.

Updates to the go-sdk and Window Destktop Edge are needed to submit the new "WindowsServer" values in posture responses.

PR #425, opened 2020-11-19
fixes #424 allow non-admins to list service terminators

### LINKED ISSUE #424: Allow authenticated clients to query service terminators
Allow authenticated clients that are non-admin to query the terminators of the services they have access to.

andrewpmartinez:
Feature request by one @ekoby


andrewpmartinez:
@plorenz let me know if you can think of any security reasons to not allow clients to query the terminators to services they have access to. I looked at the data defined in the swagger and couldn't think of any attack vector they enable.

ekoby (approved):

plorenz (changes requested):
I think we should only expose terminator id, service id and identity. We don't necessarily want to expose what router the terminator is on or the address. Exposing the binding is probably ok, but also probably not useful. The terminator address can contain information about the internals of customer networks, or in the case of SDK bound terminators, it will include their session ids, which we shouldn't expose.

ekoby:
> I think we should only expose terminator id, service id and identity. We don't necessarily want to expose what router the terminator is on or the address. Exposing the binding is probably ok, but also probably not useful. The terminator address can contain information about the internals of customer networks, or in the case of SDK bound terminators, it will include their session ids, which we shouldn't expose.

Router information (id) could be useful -- it could inform router selection on the client endpoint

andrewpmartinez:
> I think we should only expose terminator id, service id and identity. We don't necessarily want to expose what router the terminator is on or the address. Exposing the binding is probably ok, but also probably not useful. The terminator address can contain information about the internals of customer networks, or in the case of SDK bound terminators, it will include their session ids, which we shouldn't expose.

Example output for a non-admin client:
```
        {
            "_links": {
                "self": {
                    "href": "./terminators/Y97G"
                }
            },
            "createdAt": "2020-11-20T14:20:27.914Z",
            "id": "Y97G",
            "tags": {},
            "updatedAt": "2020-11-20T14:20:27.914Z",
            "identity": "",
            "routerId": "002",
            "service": {
                "_links": {
                    "configs": {
                        "href": "./services/_ZTEULTGR/configs"
                    },
                    "self": {
                        "href": "./services/_ZTEULTGR"
                    },
                    "service-edge-router-policies": {
                        "href": "./services/_ZTEULTGR/service-edge-router-policies"
                    },
                    "service-policies": {
                        "href": "./services/_ZTEULTGR/service-policies"
                    },
                    "terminators": {
                        "href": "./services/_ZTEULTGR/terminators"
                    }
                },
                "entity": "services",
                "id": "_ZTEULTGR",
                "name": "net-cat2"
            },
            "serviceId": "_ZTEULTGR"
        }        {
            "_links": {
                "self": {
                    "href": "./terminators/Y97G"
                }
            },
            "createdAt": "2020-11-20T14:20:27.914Z",
            "id": "Y97G",
            "tags": {},
            "updatedAt": "2020-11-20T14:20:27.914Z",
            "identity": "",
            "routerId": "002",
            "service": {
                "_links": {
                    "configs": {
                        "href": "./services/_ZTEULTGR/configs"
                    },
                    "self": {
                        "href": "./services/_ZTEULTGR"
                    },
                    "service-edge-router-policies": {
                        "href": "./services/_ZTEULTGR/service-edge-router-policies"
                    },
                    "service-policies": {
                        "href": "./services/_ZTEULTGR/service-policies"
                    },
                    "terminators": {
                        "href": "./services/_ZTEULTGR/terminators"
                    }
                },
                "entity": "services",
                "id": "_ZTEULTGR",
                "name": "net-cat2"
            },
            "serviceId": "_ZTEULTGR"
        }
```

ekoby:
is adding `service._links` or `service` useful at all? since it includes `serviceId` and the list is limited to specific service already

andrewpmartinez:
> is adding `service._links` or `service` useful at all? since it includes `serviceId` and the list is limited to specific service already
> 
> @andrewpmartinez that stanza will be repeated with each terminator, we should optimize it a bit

`_links` is part of the HAL standard, it will be repeated in some cases. The reason they exist is for discoverability and navigability. Some frameworks will use `_links`. Naive users (including me) use it to navigate around the API in REST tools. I'm not really inclined to take it out or try to optimize it.

ekoby:
> > is adding `service._links` or `service` useful at all? since it includes `serviceId` and the list is limited to specific service already
> > @andrewpmartinez that stanza will be repeated with each terminator, we should optimize it a bit
> 
> `_links` is part of the HAL standard, it will be repeated in some cases. The reason they exist is for discoverability and navigability. Some frameworks will use `_links`. Naive users (including me) use it to navigate around the API in REST tools. I'm not really inclined to take it out or try to optimize it.

can we take the `service` block out? it is not needed

plorenz:
> > I think we should only expose terminator id, service id and identity. We don't necessarily want to expose what router the terminator is on or the address. Exposing the binding is probably ok, but also probably not useful. The terminator address can contain information about the internals of customer networks, or in the case of SDK bound terminators, it will include their session ids, which we shouldn't expose.
> 
> Router information (id) could be useful -- it could inform router selection on the client endpoint

I would be inclined to say we should leave router out until we decide to use it. The only way I could see it being useful in selection would be if you could use that router directly, to minimize pathing. However, that might not be fastest path, so I'm not sure if you'd know when it was actually a good idea.

plorenz (approved):

PR #427, opened 2020-11-20
use custom alphabet for id generation

andrewpmartinez:
Using underscores causes issues with common name fields as they cannot
have underscores.


ekoby:
feels like it belongs in foundation

ekoby (approved):

plorenz (approved):

PR #445, opened 2020-12-10
Support router provided connection ids

plorenz:


tests/accept_manual_start_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/michaelquigley/pfxlog"
+	"github.com/openziti/fabric/controller/xt_ha"
+	"github.com/openziti/sdk-golang/ziti"
+	"github.com/openziti/sdk-golang/ziti/edge"
+	"github.com/pkg/errors"
+	"testing"
+	"time"
+)
+
+func Test_ManualStart(t *testing.T) {
```
andrewpmartinez, line 31: I think we are to the point where we have four levels of tests:

1) unit tests
2) internal integration tests
3) API integration tests
4) Connection oriented integration tests

1 and 2 execute together as test files next to the source code.

3 and 4 execute as `apitests`. I feel like they should be split. I am not asking this to be done here, but it is a thought I've had as the category of 4 gets larger.

andrewpmartinez (approved):

tests/accept_manual_start_test.go
plorenz: seems like a reasonable distinction. Maybe apitests and dataflow as tags.

router/xgress_edge/accept.go
```diff
@@ -45,6 +46,7 @@ func (handler edgeBindHandler) BindChannel(ch channel2.Channel) error {
 		listener:     handler.listener,
 		fingerprints: fpg.FromCerts(ch.Certificates()),
 		ch:           ch,
+		idSeq:        math.MaxUint32 / 2,
```
andrewpmartinez, line 49: For curiosity reasons, why max int / 2?

plorenz: The id ranges are split into two. 0 to (maxuint32 / 2 - 1) for client generated ids and maxuint32 / 2 -> maxuint32 for server generated ids. This way we don't run into id conflicts.
Eugene pointed out that http2 does something similar, but they use even number for one and odd numbers for the other.

PR #452, opened 2021-01-06
adds support of API Session Certificates

andrewpmartinez:
- adds the ability for any authenticated API Session to create ephemeral
 certificates
- allows UPDB session connection
- adds create/delete/read of current API session certificates
- add session cert PEM result on create and detail
- fixes ER not updating sessions with new certs from API session
- fixes naming of session vs. apiSession on members, functions, etc
- organizes ER state manager functions into apiSession vs. session
- adds API session ids to API sessions and sessions sent from the
  controller to ERs
- ensure API and ns fingerprints are the same
- filter API session certs by valid periods
- clarify fingerprint func naming


pb/edge_ctrl_pb/edge_ctrl.proto
```diff
@@ -54,11 +54,13 @@ message Session {
...
     Service service = 5;
     string id = 7;
+    string apiSessionId = 8;
 }
 
 message ApiSession {
     string token = 1;
     repeated string certFingerprints = 2;
+    string id = 3;
 }
```
andrewpmartinez, line 64: These allow sessions and API Sessions to easily be linked in the router's cache.

andrewpmartinez:
I have no idea why I can't add some people to this review. They don't show up in the reviewer list when blank or when I enter their usernames.

rentallect (approved):

router/internal/fabric/manager.go
```diff
@@ -83,32 +89,44 @@ func GetStateManager() StateManager {
...
-	sm.sessionsByToken.Store(session.Token, session)
+	sm.apiSessionsByToken.Store(session.Token, session)
 	sm.Emit(EventAddedSession, session)
 }
 
 func (sm *StateManagerImpl) UpdateApiSession(session *edge_ctrl_pb.ApiSession) {
 	pfxlog.Logger().Debugf("updating session [%s] fingerprints [%s]", session.Token, session.CertFingerprints)
-	sm.sessionsByToken.Store(session.Token, session)
+	sm.apiSessionsByToken.Store(session.Token, session)
+
```
plorenz, line 111: I don't know the frequency of this is, but would it make sense to have a second id -> token map, so that we could do this with two map lookups, rather than having to scan the map?

plorenz (approved):

plorenz (approved):

PR #454, opened 2021-01-08
Add health check implementation for tunnelers and add default hosting precedence/cost to identity

plorenz:


.github/workflows/main.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         uses: netfoundry/ziti-ci@v1
 
       - name: Build and Test
-        run: go test ./... --tags apitests
+        run: go test ./... --tags apitests,slowtests
```
andrewpmartinez, line 27: oooh slowtests!

andrewpmartinez (approved):

controller/internal/routes/identity_api_model.go
```diff
@@ -59,18 +60,27 @@ func (factory *IdentityLinkFactoryImpl) Links(entity models.Entity) rest_model.L
...
-		IdentityTypeId: identityTypeId,
-		IsDefaultAdmin: false,
-		IsAdmin:        *identity.IsAdmin,
-		RoleAttributes: identity.RoleAttributes,
+		Name:                     stringz.OrEmpty(identity.Name),
+		IdentityTypeId:           identityTypeId,
+		IsDefaultAdmin:           false,
+		IsAdmin:                  *identity.IsAdmin,
+		RoleAttributes:           identity.RoleAttributes,
+		DefaultHostingPrecedence: ziti.GetPrecedenceForLabel(string(identity.DefaultHostingPrecedence)),
```
ekoby, line 82: Why? Precedence -> string -> Precedence?

plorenz: identity.DefaultHostingPrecedence is of type TerminatorPrecedence which is generated type, but is just a type alias for string.

So we're really just going string -> Precedence

PR #459, opened 2021-01-14
Implement service change notifications

plorenz:


tests/model_sdk_performance_test.go
```diff
@@ -28,6 +28,21 @@ type modelPerf struct {
 	*TestContext
 }
 
+func Test_SpecRamco(t *testing.T) {
```
andrewpmartinez, line 31: Probably shouldn't put this company name here.

andrewpmartinez (approved):

tests/model_sdk_performance_test.go
plorenz: ah, you are correct, will fix

PR #475, opened 2021-01-19
Db validate guardrails

plorenz:


controller/internal/routes/database_router.go
```diff
@@ -68,27 +89,43 @@ func (r *DatabaseRouter) CreateSnapshot(ae *env.AppEnv, rc *response.RequestCont
...
-	errorHandler := func(err error, fixed bool) {
-		description := err.Error()
-		results = append(results, &rest_model.DataIntegrityCheckDetail{
-			Description: &description,
-			Fixed:       &fixed,
-		})
+	if r.integrityCheck.running.CompareAndSwap(false, true) {
+		r.integrityCheck.fixingErrors = fixErrors
+		go r.runDataIntegrityCheck(ae, fixErrors)
+		rc.Respond(&rest_model.Empty{Data: map[string]interface{}{}, Meta: &rest_model.Meta{}}, http.StatusOK)
```
andrewpmartinez, line 95: `rc.RespondWithEmptyOk()` is a convenience function.

specs/swagger.yml
```diff
@@ -3060,28 +3060,46 @@ paths:
         '429':
           $ref: '#/responses/rateLimitedResponse'
   /database/check-data-integrity:
-    get:
-      summary: Runs an data integrity scan on the datastore and returns any found issues
-      description: Runs an data integrity scan on the datastore and returns any found issues. Requires admin access.
+    post:
+      summary: Starts a data integrity scan on the datastore
+      description: Starts a data integrity scan on the datastore. Requires admin access.
```
andrewpmartinez, line 3065: I would mention what causes `429`. Same w/ `/database/fix-data-integrity` below. 

Do `check` and `fix` limit each other too? If so that is worth mentioning.

Example:

> Starts a data integrity scan on the data store. Requires admin access, can only run one instance.

andrewpmartinez (approved):

controller/internal/routes/database_router.go
plorenz: That was supposed to be StatusAccepted, which is why I didn't use the empty ok func. Fixed.

specs/swagger.yml
plorenz: :+1:  Added more doc

PR #492, opened 2021-01-27
Fix close stall by moving notify after seq close

plorenz:


router/xgress_edge/fabric.go
```diff
@@ -188,14 +188,6 @@ func (conn *edgeXgressConn) close(notify bool, reason string) {
 
 	log := pfxlog.ContextLogger(conn.Channel.Label()).WithField("connId", conn.Id())
 	log.Debugf("closing message sink, reason: %v", reason)
```
dovholuknf, line 190: should this message travel with this block?

plorenz: no, it's supposed to be at the top. I should fix the text, though

plorenz: fixed

dovholuknf (approved):

PR #496, opened 2021-02-01
limit public query options to 500 whe "none" is specified

andrewpmartinez:


controller/internal/routes/query_options.go
```diff
@@ -31,46 +31,22 @@ const (
...
 
-func (qo *QueryOptions) String() string {
+func (qo *PublicQueryOptions) String() string {
 	if qo == nil {
 		return "nil"
 	}
 	return fmt.Sprintf("[QueryOption Predicate: '%v', Sort: '%v', Paging: '%v']", qo.Predicate, qo.Sort, qo.Paging)
 }
 
-func (qo *QueryOptions) ValidateAndCorrect() {
```
andrewpmartinez, line 47: This was long longer called. Most of the logic is handled in the query processing logic.

controller/internal/routes/query_options.go
```diff
@@ -31,46 +31,22 @@ const (
 	OffsetDefault = 0
 )
 
-type QueryOptions struct {
+// Represents external query options, which limits some query options (i.e. return all vs max limit). PublicQueryOptions
+// can be used internally as long as the public option limitations are fitting.
+type PublicQueryOptions struct {
```
andrewpmartinez, line 36: Renamed so it is clear that this struct is used for public API responses (currently only the REST API)

dovholuknf (approved):

PR #498, opened 2021-02-01
fixes #497 send session header

### LINKED ISSUE #497: Add server response header
for example:
`server: ziti-controller/v0.18.3`

this will allow SDK to detect controller upgrades without separate querying and adjust/reset supported features

andrewpmartinez:
- add server headers during HTTP handler delegation rather than at HTTP
  response time
- remove all old header setting locations
- remove unused status writer and other response code
- add manual set on authentication for session headers


controller/server/controller.go
```diff
@@ -261,6 +262,9 @@ func (c *Controller) Run() {
 				return
 			}
 
+			//after request context is filled so that api session is present for session expiration headers
+			response.AddHeaders(rc)
```
andrewpmartinez, line 266: This location is the root of the HTTP server handler delegation. Setting the headers here means unless downstream code unsets or resets it, these headers will appear on every response. This is an improvement as previously headers were set in multiple locations. 

The only additional place that these headers need to be set is during authentication when the ApiSession used for expiration headers comes into existence after this code runs.

ekoby (approved):

plorenz (approved):

PR #499, opened 2021-02-01
edge router enhancements for browZer support

rentallect:


pb/edge_ctrl_pb/edge_ctrl.proto
```diff
@@ -33,6 +33,7 @@ message ClientHello {
     string hostname = 2;
     repeated string protocols = 3;
     map<string, string> data = 4;
+    repeated string protoports = 5;
```
andrewpmartinez, line 36: We can use camel case here and use the full name.

`protocolPorts`

router/internal/router/config.go
```diff
@@ -248,6 +253,12 @@ func (config *Config) loadListener(rootConfigMap map[interface{}]interface{}) er
 					return errors.New("required value [listeners.edge.options.advertise] was not a string or was not found")
 				}
 
+				parts := strings.Split(advertise, ":")
+				if parts[1] != edgeListenPort {
+					msg := fmt.Sprintf("port in [listeners.edge.options.advertise] must equal port in [listeners.edge.address] but did not. Got [%s] [%s]", parts[1], edgeListenPort)
```
andrewpmartinez, line 258: The ports don't have to be equal. It is possible the advertised address is some proxy that listens on a different port.

router/xgress_edge/factory.go
```diff
@@ -53,7 +53,34 @@ func (factory *Factory) Enabled() bool {
...
+		supportedProtocols = append(supportedProtocols, "tls")
+		protoPorts = append(protoPorts, "tls:"+parts[1])
+		pfxlog.Logger().Debugf("HelloHandler will contain hostname=[%s] supportedProtocols=%v protoPorts=%v", hostname, supportedProtocols, protoPorts)
+	}
+	if factory.config.WSAdvertise != "" {
+		parts := strings.Split(factory.config.WSAdvertise, ":")
+		if hostname != "" {
+			if hostname != parts[0] {
+				msg := fmt.Sprintf("cannot have different hostnames within multiple edge binding.advertise; got [%s] [%s]", hostname, parts[0])
+				panic(msg)
```
andrewpmartinez, line 74: I think this should be a `pfxlog.Logger.Fatalf(...)`

router/xgress_edge_transport/connection.go
```diff
@@ -143,7 +143,7 @@ func (c *edgeTransportXgressConn) Write(p []byte) (n int, err error) {
 		//if crypto enabled and we have not setup e2e
 		if c.cryptoCtx.rxKey != nil {
 			if len(p) != secretstream.StreamHeaderBytes {
-				return 0, fmt.Errorf("error establishing crypto: expected key length %d got %d", len(p), secretstream.StreamHeaderBytes)
+				return 0, fmt.Errorf("error establishing crypto: expected key length %d got %d", secretstream.StreamHeaderBytes, len(p))
```
andrewpmartinez, line 146: These values look swapped.

specs/swagger.yml
```diff
@@ -1716,7 +1716,7 @@ paths:
         must present a client certificate signed by CA associated with the enrollment. This enrollment is similar to
         CA auto enrollment except that is required the identity to be pre-created.
 
-        As the client certificat has been pre-exchanged there is no CSR input to this enrollment method.
+        As the client certificate has been pre-exchanged there is no CSR input to this enrollment method.
```
andrewpmartinez, line 1719: THANK YOU!

router/internal/router/config.go
```diff
@@ -229,7 +234,7 @@ func (config *Config) loadListener(rootConfigMap map[interface{}]interface{}) er
 		}
 	}
 
-	if (edgeBinding == nil) && (edgeWssBinding == nil) {
+	if (edgeBinding == nil) && (edgeWsBinding == nil) {
 		return errors.New("required binding [edge] not found in [listeners]")
```
andrewpmartinez, line 238: I would augment this message to say

`required binding [edge] not found in [listeners], at least one edge binding is required`

andrewpmartinez (changes requested):
Looks good some small things to change.

pb/edge_ctrl_pb/edge_ctrl.proto
rentallect: I changed 'protoports' to 'protocolPorts' everywhere

router/internal/router/config.go
rentallect: Fair enough, but I will emit a WARN log msg here to help call out cases where someone might have typo'd (like I have done).

router/xgress_edge/factory.go
rentallect: done

router/xgress_edge_transport/connection.go
rentallect: No, it is correct as I have it.  We expect secretstream.StreamHeaderBytes (which is defined here: https://github.com/golang/crypto/blob/master/chacha20poly1305/chacha20poly1305.go#L28))

It was actually swapped before this.  We witnessed this mis-formatting issue in the field a while back, and this fix has been pending in my branch.

router/internal/router/config.go
rentallect: done

specs/swagger.yml
rentallect: 👍

plorenz (approved):

PR #533, opened 2021-02-12
Validate sessions on controller instead of edge router

plorenz:


pb/edge_ctrl_pb/edge_ctrl.proto
```diff
@@ -3,182 +3,240 @@ syntax = "proto3";
...
-            Android = 0;
-            iOS = 10;
-            Linux = 20;
-            macOS = 30;
-            Windows = 40;
-            WindowsServer = 50;
-        }
-    }
+message CreateCircuitRequest {
+  string sessionToken = 1;
```
andrewpmartinez, line 196: this should include `sessionId` for logging purpose.

andrewpmartinez, line 196: This message should include `sessionId` to be used on the receiving end for logging.

pb/edge_ctrl_pb/edge_ctrl.proto
```diff
@@ -3,182 +3,240 @@ syntax = "proto3";
...
+  Failed = 2;
+}
+
+message CreateTerminatorRequest {
+  string sessionToken = 1;
+  repeated string fingerprints = 2;
+  map<uint32, bytes> peerData = 3;
+  uint32 cost = 4;
+  TerminatorPrecedence precedence = 5;
+  string identity = 6;
```
andrewpmartinez, line 220: is this `identityId` ?

pb/edge_ctrl_pb/edge_ctrl.proto
```diff
@@ -3,182 +3,240 @@ syntax = "proto3";
...
+}
+
+enum TerminatorPrecedence {
+  Default = 0;
+  Required = 1;
+  Failed = 2;
+}
+
+message CreateTerminatorRequest {
+  string sessionToken = 1;
```
andrewpmartinez, line 215: This message should include `sessionId` to be used on the receiving end for logging.

pb/edge_ctrl_pb/edge_ctrl.proto
```diff
@@ -3,182 +3,240 @@ syntax = "proto3";
...
+  repeated string fingerprints = 2;
+  map<uint32, bytes> peerData = 3;
+  uint32 cost = 4;
+  TerminatorPrecedence precedence = 5;
+  string identity = 6;
+  bytes identitySecret = 7;
+}
+
+message RemoveTerminatorRequest {
+  string sessionToken = 1;
```
andrewpmartinez, line 225: This message should include `sessionId` for logging.

pb/edge_ctrl_pb/edge_ctrl.proto
```diff
@@ -3,182 +3,240 @@ syntax = "proto3";
...
+  repeated string fingerprints = 2;
+  string terminatorId = 3;
+}
+
+message UpdateTerminatorRequest {
+  string sessionToken = 1;
```
andrewpmartinez, line 231: This message should include sessionId for logging.

pb/edge_ctrl_pb/edge_ctrl.proto
```diff
@@ -3,182 +3,240 @@ syntax = "proto3";
...
+  repeated string fingerprints = 2;
+  string terminatorId = 3;
+  uint32 cost = 4;
+  TerminatorPrecedence precedence = 5;
+  bool updatePrecedence = 6;
+  bool updateCost = 7;
 }
 
+message ValidateSessionsRequest {
+  repeated string sessionTokens = 1;
```
andrewpmartinez, line 241: This should include `sessionId` for logging purposes

controller/handler_edge_ctrl/common.go
```diff
@@ -0,0 +1,229 @@
...
+	return self.ch
+}
+
+func (self *baseRequestHandler) returnError(ctx requestContext, err error) {
+	responseMsg := channel2.NewMessage(int32(edge_ctrl_pb.ContentType_ErrorType), []byte(err.Error()))
+	responseMsg.ReplyTo(ctx.GetMessage())
+	if err = self.ch.Send(responseMsg); err != nil {
+		pfxlog.
+			ContextLogger(self.ch.Label()).
+			WithError(err).
```
andrewpmartinez, line 47: Should log `sessionId`

controller/handler_edge_ctrl/common.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+	return true
+}
+
+func (self *baseRequestContext) loadSession(token string) {
+	if self.err == nil {
+		self.session, self.err = self.handler.getAppEnv().Handlers.Session.ReadByToken(token)
+		if self.err != nil {
+			logrus.
+				WithField("token", token).
```
andrewpmartinez, line 96: Should log `sessionId`

controller/handler_edge_ctrl/common.go
```diff
@@ -0,0 +1,229 @@
...
+				WithError(self.err).Errorf("invalid session")
+		}
+	}
+}
+
+func (self *baseRequestContext) checkSessionType(sessionType string) {
+	if self.err == nil {
+		if self.session.Type != sessionType {
+			logrus.
+				WithField("token", self.session.Token).
```
andrewpmartinez, line 107: Should log `sessionId`


I've added similar messages in other places. I'm not going to add any more, but where we can output ids we should. If the id isn't present but it could be, we should add it.

controller/handler_edge_ctrl/create_terminator.go
```diff
@@ -0,0 +1,132 @@
...
+
+	if ctx.req.Cost > math.MaxUint16 {
+		self.returnError(ctx, errors.Errorf("invalid cost %v. cost must be between 0 and %v inclusive", ctx.req.Cost, math.MaxUint16))
+		return
+	}
+
+	terminator := &network.Terminator{
+		Service:        ctx.session.ServiceId,
+		Router:         ctx.sourceRouter.Id,
+		Binding:        "edge",
```
andrewpmartinez, line 104: Constant?

router/internal/edgerouter/config.go
```diff
@@ -14,36 +14,43 @@
...
 const (
-	DefaultHeartbeatIntervalSeconds = 60
-	MinHeartbeatIntervalSeconds     = 10
-	MaxHeartbeatIntervalSeconds     = 60
+	DefaultHeartbeatIntervalSeconds   = 60
+	MinHeartbeatIntervalSeconds       = 10
+	MaxHeartbeatIntervalSeconds       = 60
+	DefaultSessionValidateChunkSize   = 1000
+	DefaultSessionValidateMinInterval = "250ms"
+	DefaultSessionValidateMaxInterval = "1500ms"
```
andrewpmartinez, line 39: Should include a postfix unit? Example `DefaultSessionValidateMinIntervalMilliseconds. Also not a string? 

We don't follow this pattern anywhere else in the edge right now.

controller/handler_edge_ctrl/common.go
```diff
@@ -0,0 +1,229 @@
...
+			self.err = errors.New("invalid session")
+		}
+	}
+}
+
+func (self *baseRequestContext) checkSessionFingerprints(fingerprints []string) {
+	if self.err == nil {
+		now := time.Now()
+
+		var sessionCerts []string
```
andrewpmartinez, line 119: `sessionFingerprints` would be a better name. If you swap this loop with the loop below you cans just reuse the `prints` slice.

andrewpmartinez (changes requested):
I didn't add all of the comments, but we should strive to output `id`s instead of `tokens` where possible. 

Per recent debugging sessions with unstable networks, having `ids` output instead of `tokens` is much easier to deal with and do quick look ups via the API.

Per recent security discussions, tokens should not be output. They are only part of a secret but it is another layer of security. We should strive to not output `tokens`. While `API Session Tokens` are more valuable `Session Tokens` should be treated with respect as well as they are considered a secret.

pb/edge_ctrl_pb/edge_ctrl.proto
plorenz: It's not available from the edge router to send.

pb/edge_ctrl_pb/edge_ctrl.proto
plorenz: No it's a fabric terminator identifier used for grouping terminators. Maybe renamed to group?

pb/edge_ctrl_pb/edge_ctrl.proto
plorenz: session id not available here

controller/handler_edge_ctrl/create_terminator.go
plorenz: Added a edge_common.Binding constant and used where appropriate

router/internal/edgerouter/config.go
plorenz: We've started using this in the fabric. It's shorter, more flexible and uses the built-in time.ParseDuration method. It's a relatively standard way of specifying intervals/durations. If you're strongly against it, we should talk it through.

controller/handler_edge_ctrl/common.go
plorenz: Fixed

plorenz:
I've switching sessionToken with sessionId where I can on the server side. Will leave the hashing for a future pass.

router/xgress_edge/fabric.go
```diff
@@ -160,62 +163,66 @@ func (conn *edgeXgressConn) WritePayload(p []byte, headers map[uint8][]byte) (n
...
 	}
+
+	if self.onClose != nil {
+		self.onClose()
+	}
 }
 
-func (conn *edgeXgressConn) Accept(msg *channel2.Message) {
-	if err := conn.seq.Push(msg); err != nil {
+func (self *edgeXgressConn) Accept(msg *channel2.Message) {
```
andrewpmartinez, line 219: `self` reminds me of C#. I like this convention personaly.

andrewpmartinez (approved):

PR #536, opened 2021-02-15
Don't use half close with udp connections

plorenz:


tunnel/tunnel.go
```diff
@@ -69,16 +69,18 @@ func Run(zitiConn net.Conn, clientConn net.Conn) {
...
 		"dst-local": dst.LocalAddr(), "dst-remote": dst.RemoteAddr()}
 
 	logger := log.WithFields(loggerFields)
 	defer func() {
-		if cw, ok := dst.(edge.CloseWriter); ok {
+		if cw, ok := dst.(edge.CloseWriter); halfClose && ok {
+			logger.Debug("doing half-close")
 			_ = cw.CloseWrite()
 		} else {
+			logger.Debug("doing half-close")
```
scareything, line 83: should log "doing full close"

scareything (approved):

tunnel/tunnel.go
plorenz: fixed, thank you

PR #540, opened 2021-02-17
do not remove api sessions added after/during sync

andrewpmartinez:


plorenz (approved):

controller/handler_edge_ctrl/heartbeat.go
```diff
@@ -38,17 +39,35 @@ func (h *sessionHeartbeatHandler) ContentType() int32 {
 	return env.ApiSessionHeartbeatType
 }
 
-func (h *sessionHeartbeatHandler) HandleReceive(msg *channel2.Message, _ channel2.Channel) {
+func (h *sessionHeartbeatHandler) HandleReceive(msg *channel2.Message, ch channel2.Channel) {
```
andrewpmartinez, line 42: @plorenz I added some heartbeat invalid api token logic here. You did your review before I pushed this.

controller/handler_edge_ctrl/heartbeat.go
```diff
@@ -38,17 +39,35 @@ func (h *sessionHeartbeatHandler) ContentType() int32 {
...
+					Debugf("api session tokens not found during heartbeat, sending delete: %v", notFoundTokens)
+
+				msgStruct := &edge_ctrl_pb.ApiSessionRemoved{
+					Tokens: notFoundTokens,
+				}
+
+				msgBytes, _ := proto.Marshal(msgStruct)
+				msg := channel2.NewMessage(env.ApiSessionRemovedType, msgBytes)
+
+				//if we don't delete it this time, we will try again when we see it
```
plorenz, line 69: I would just use Send. The timeout only adds overhead if you're not checking for an error

andrewpmartinez: I was trying to avoid blocking as this is an unbounded go routine generator.

plorenz: Send will return faster than SendWithTimeout. Send will return as soon as it's queued. SendWithTimeout will wait until it's off the queue and written to network, timing out if it doesn't make it onto the network in time. 

SendWithTimeout won't start the timeout until the message is queued for send.

PR #554, opened 2021-02-25
Fix update terminator. Add test for terminator updates. Remove unused…

plorenz:
… store_events


router/xgress_edge/listener.go
```diff
@@ -292,6 +292,8 @@ func (self *edgeClientConn) processUpdateBind(req *channel2.Message, ch channel2
 	}
 
 	request := &edge_ctrl_pb.UpdateTerminatorRequest{
```
plorenz, line 294: This is the actual fix. Everything else is just tests or removal of dead code

andrewpmartinez (approved):

PR #559, opened 2021-03-01
Add session route to Session/id/route-path

andrewpmartinez:


controller/internal/routes/session_api_model.go
```diff
@@ -83,6 +83,19 @@ func MapSessionToRestModel(ae *env.AppEnv, sessionModel *model.Session) (*rest_m
 		return nil, err
 	}
```
plorenz, line 85: This can get expensive. If we're listing lots of edge sessions, we're going to iterations potentially thousands of fabric sessions each time. Fabric sessions will also likely be persisted at some point in the future, which will make this more expensive. Is there a way to make this an optional field? Or maybe a separate API? 

This would be easy to do in the ziti CLI if we had a REST fabric api. Someday...

Also there maybe 0-N fabric sessions for each edge session, here we're only show at most 1, which may be confuse the viewer.

andrewpmartinez: Expensive: yea it is at scale.

If there was a way to do the lookup by `ClientId` that would be faster or if `ClientId` and `Session.Id` were the same value by convention. IIRC `ClientId` and `Session.Id` were the same long ago but aren't now. I don't know the exact reasons, I assume it is because of the way connections happen now and a single edge session can generate multiple fabric sessions.

What are your thoughts on adding a map inside of the fabric indexed by `ClientId` and maintained alongside the one indexed by fabric session id?


Right now we have 

```
type sessionController struct {
	sessions          cmap.ConcurrentMap // map[string]*Session
	sessionsByService cmap.ConcurrentMap // map[string]*Sessions
}
```

This could become:

```
type sessionController struct {
	sessions          cmap.ConcurrentMap // map[string]*Session
	sessionsByService cmap.ConcurrentMap // map[string]*Sessions
	sessionsByClientId cmap.ConcurrentMap // map[string]*Sessions
}
```

Re: Fabric API

Yes that is useful but any tooling would have to aggregate the data to show any kind of list. Making coordinated calls like that is generally avoided. I do agree a similar output would be in the Fabric API.

Re: Separate Call

That is possible but also hurts listing. I actually implemented it this way at first.

plorenz: I'm not sure if there's any reason it couldn't be switched back to sessionId. I switched it to session id accidentally as part of the session verification refactor and it broke reporting, but nothing else. So as long as we keep the reporting folks in the loop, we can switch it back.

Do we need this data to show up in the list? Seems like the kind of thing you'd be checking per edge session. If we can have a separate call that just does the loop for a given edge session, we can let it be somewhat inefficient. 

I don't mind adding an extra map, but it would need to be a multi-map since you can have multiple sessions per session. (I really would like to see fabric sessions renamed to something else, like circuit)

andrewpmartinez: All right I'm just going to move it to a separate endpoint.

plorenz (approved):

PR #573, opened 2021-03-08
adds thread safe router state

andrewpmartinez:
Router state used to be stored and accessed in a thread unsafe way.
State is now locks on a mutex for setting values and values are copied
out for referencing.

### DIFF
--- a/controller/env/broker.go
+++ b/controller/env/broker.go
@@ -180,8 +180,8 @@ func (broker *Broker) apiSessionCertificateHandler(args ...interface{}) {
 }
 
 func (broker *Broker) IsEdgeRouterOnline(id string) bool {
-	status := broker.GetEdgeRouterState(id)
-	return status.IsOnline
+	state := broker.GetEdgeRouterState(id)
+	return state.IsOnline
 }
 
 func (broker *Broker) GetEdgeRouterState(id string) RouterStateValues {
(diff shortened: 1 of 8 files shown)


plorenz:
I think it's probably OK the way it is. It looks like it's still reading w/out locks. Since there's probably more reads than writes, and writes should be fairly infrequent, simplest way might be to use an immutable state object and just replace the whole object on change.

plorenz (approved):

andrewpmartinez:
>It looks like it's still reading w/out locks.

The way reading works is by calling Values() and getting a copy of the state at that instant. After that, the reads don't matter because it was a copy.

plorenz:
> > It looks like it's still reading w/out locks.
> 
> The way reading works is by calling Values() and getting a copy of the state at that instant. After that, the reads don't matter because it was a copy.
> 
> Most of the reads are done by API model code which doesn't hold on to them or use them for very long.

That's what I thought it would do, but Values() returns the internal state, not a copy of the internal state:

andrewpmartinez:
> > > It looks like it's still reading w/out locks.
> > 
> > 
> > The way reading works is by calling Values() and getting a copy of the state at that instant. After that, the reads don't matter because it was a copy.
> > Most of the reads are done by API model code which doesn't hold on to them or use them for very long.
> 
> That's what I thought it would do, but Values() returns the internal state, not a copy of the internal state:

And the return type isn't a pointer. So it will return by value.

andrewpmartinez:
I did just notice the map is by reference. I am fixing that now.

PR #582, opened 2021-03-12
Pass FieldChecker to toBoltEntityForPatch. Add some ReadByName method…

plorenz:
…s. Move some code to allow re-use

### DIFF
--- a/controller/model/identity_handlers.go
+++ b/controller/model/identity_handlers.go
@@ -207,6 +207,15 @@ func (handler *IdentityHandler) Read(id string) (*Identity, error) {
 	return entity, nil
 }
 
+func (handler *IdentityHandler) ReadByName(name string) (*Identity, error) {
+	entity := &Identity{}
+	nameIndex := handler.env.GetStores().Identity.GetNameIndex()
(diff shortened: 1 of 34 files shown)


controller/model/identity_handlers.go
```diff
@@ -207,6 +207,15 @@ func (handler *IdentityHandler) Read(id string) (*Identity, error) {
 	return entity, nil
 }
 
+func (handler *IdentityHandler) ReadByName(name string) (*Identity, error) {
```
andrewpmartinez, line 210: It isn't immediately apparent why we need `ReadByName` for identity and edge service. I believe config types are referenced by name by REST API Clients, so that doesn't seem strange to me. Config type names are also relatively static.

I don't see the identity/service ones used and thus I don't understand why we need it. I want to ensure we are avoiding relying on names other than for initial lookups as names can be changed.

plorenz: These will be used by tunnel control channel handlers and the router identities code. Services are dialed by name, so we need to be able to look them up by name. Router identities will be linked to routers by name, so we look them up by name and when we're modifying names to keep them in sync, we can look them up by the old name.

andrewpmartinez (approved):

PR #585, opened 2021-03-15
use batch for api access last updates

andrewpmartinez:


controller/model/api_session_handlers.go
```diff
@@ -92,8 +93,29 @@ func (handler *ApiSessionHandler) Delete(id string) error {
...
 }
 
-// MarkActivity returns tokens that were not found if any and/or an error.
-func (handler *ApiSessionHandler) MarkActivity(tokens []string) ([]string, error) {
+func (handler *ApiSessionHandler) MarkActivityById(apiSessionId string) {
+	err := handler.GetDb().Batch(func(tx *bbolt.Tx) error {
+		store := handler.Store.(persistence.ApiSessionStore)
+		mutCtx := boltz.NewMutateContext(tx)
+
+		apiSession, err := store.LoadOneById(tx, apiSessionId)
```
andrewpmartinez, line 101: At this level, we need persistence.ApiSession no a model.ApiSession hence why id is taken.

andrewpmartinez: mmm, might be able to just use a mock persistence.ApiSession

plorenz (approved):

PR #586, opened 2021-03-15
buffer lastUpdatedAt in memory, flush on interval/shutdown

andrewpmartinez:
Needs a foundation fix that removes an os.Exit() call from another
signal listener.

Should lower throughput for disk I/O as multiple lastUpdatedAt triggers
can happen in period of time via edge routers and API calls.


controller/internal/policy/session.go
```diff
@@ -59,7 +59,10 @@ func (s *SessionEnforcer) Run() error {
 	for i := 0; i < maxIterations; i++ {
 		ids := make([]string, 0, maxDeletePerIteration)
 		err := s.appEnv.GetHandlers().ApiSession.StreamIds(query, func(id string, err error) error {
-			ids = append(ids, id)
+			if lastUpdateAt, hasUnflushedValue := s.appEnv.GetHandlers().ApiSession.HeartbeatCollector.LastUpdatedAt(id); !hasUnflushedValue || lastUpdateAt.Before(oldest) {
```
andrewpmartinez, line 62: For API Session, deletion use the values from memory over bbolts in case we are between a flush and at the edge of an API Session's life.

controller/model/api_session_heartbeats.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (self *HeartbeatCollector) Start() {
+	go self.run()
+}
+
+func (self *HeartbeatCollector) run() {
+	for {
+		select {
+		case <-self.closeNotify:
+			self.flush() //flush on stop
```
andrewpmartinez, line 88: On shutdown, flush. Graceful shutdowns need to work and hence the other shutdown work in this PR.

controller/model/api_session_heartbeats.go
```diff
@@ -0,0 +1,143 @@
...
+			return
+		case <-time.After(self.flushInterval):
+			self.flush()
+		}
+	}
+}
+
+func (self *HeartbeatCollector) Stop() {
+	close(self.closeNotify)
```
andrewpmartinez, line 97: I wish we had a single context or closeNotify channel.

controller/model/api_session_heartbeats.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (self *HeartbeatCollector) flush() {
+	if self.isFlushing.CompareAndSwap(false, true) {
+		pfxlog.Logger().Trace("flushing heartbeat collector")
+
+		var ids []string
+
+		self.apiSessionLastUpdate.IterCb(func(key string, v interface{}) {
+			if len(ids) >= self.flushSize {
```
andrewpmartinez, line 108: Flush in batches so we don't blow out a transaction.

controller/server/controller.go
```diff
@@ -349,34 +350,43 @@ func (c *Controller) RunAndWait() {
...
 	c.AppEnv.Broker.Stop()
 
-	pfxlog.Logger().Info("edge controller shutting down")
+	c.AppEnv.GetHandlers().ApiSession.HeartbeatCollector.Stop()
+
+	pfxlog.Logger().Info("edge controller: stopped")
+
+	pfxlog.Logger().Info("fabric controller: shutting down...")
+
+	c.AppEnv.GetHostController().Shutdown()
```
andrewpmartinez, line 383: This shuts down the fabric controller gracefully. For some time we haven't been and that the process running w/o calling os.Exit().

controller/model/api_session_handlers.go
```diff
@@ -33,13 +33,16 @@ func NewApiSessionHandler(env Env) *ApiSessionHandler {
 	handler := &ApiSessionHandler{
 		baseHandler: newBaseHandler(env, env.GetStores().ApiSession),
 	}
+	handler.HeartbeatCollector = NewHeartbeatCollector(env,100, 2*time.Minute, handler.heartbeatFlush)
```
andrewpmartinez, line 36: Hard-coded defaults. I'll add configuration values.

controller/server/controller.go
```diff
@@ -349,34 +350,43 @@ func (c *Controller) RunAndWait() {
...
+
+	pfxlog.Logger().Info("fabric controller: stopped")
+
+	log.Info("shutdown complete")
```
andrewpmartinez, line 387: Oddly enough, sometimes this message isn't output in my testing. Sometimes it is. I didn't see a clear way to force logrus to flush.

controller/internal/routes/current_api_session_api_model.go
```diff
@@ -57,31 +57,42 @@ func (factory *CurrentApiSessionLinkFactoryImpl) Links(entity models.Entity) res
...
+			Identity:     ToEntityRef(apiSession.Identity.Name, apiSession.Identity, IdentityLinkFactory),
+			IdentityID:   &apiSession.IdentityId,
+			IPAddress:    &apiSession.IPAddress,
+			Token:        &apiSession.Token,
+			AuthQueries:  authQueries,
+			IsMfaEnabled: &isMfaEnabled,
 		},
 		ExpiresAt:         &expiresAt,
 		ExpirationSeconds: &expirationSeconds,
 	}
```
plorenz, line 87: This is different than what's above in api_session_api_model. Should they be the same? Should this only ever return one of LastActivityAt or CachedLastActivityAt

controller/model/api_session_heartbeats.go
```diff
@@ -0,0 +1,151 @@
...
+// times.
+func NewHeartbeatCollector(env Env, batchSize int, updateInterval time.Duration, action func([]*Heartbeat)) *HeartbeatCollector {
+	collector := &HeartbeatCollector{
+		apiSessionLastUpdate: cmap.New(),
+		updateInterval:       updateInterval,
+		batchSize:            batchSize,
+		flushAction:          action,
+		closeNotify:          make(chan struct{}, 0),
+	}
+
```
plorenz, line 54: Is it worth adding this? We have to handle failed updates either way.

controller/server/controller.go
```diff
@@ -349,34 +345,43 @@ func (c *Controller) RunAndWait() {
 	c.waitForShutdown()
 }
```
plorenz, line 347: remnant?

controller/internal/routes/current_api_session_api_model.go
andrewpmartinez: Fixed

controller/model/api_session_heartbeats.go
andrewpmartinez: Is it worth adding what?

plorenz: the api session event listener to remove entries from the map when session are removed. They'll get removed as they get flushed and we have to deal with write failures either way.

andrewpmartinez: Right now flushing doesn't remove them from the map. So the map can just grow.

controller/server/controller.go
andrewpmartinez: removed

controller/model/api_session_heartbeats.go
plorenz: ah, ok

plorenz (approved):

PR #600, opened 2021-03-22
set mfa posture on enroll

andrewpmartinez:
- update api session session state on mfa enroll, remove
- update posture data state on mfa enroll
- centralize setting mfa posture data
- pass isMfaRequired and isMfaComplete in apiSession/currentApiSession
- on posture data  change trigger service update

### DIFF
--- a/controller/internal/routes/current_identity_router.go
+++ b/controller/internal/routes/current_identity_router.go
@@ -18,12 +18,17 @@ package routes
 
 import (
 	"github.com/go-openapi/runtime/middleware"
+	"github.com/michaelquigley/pfxlog"
 	"github.com/openziti/edge/controller/apierror"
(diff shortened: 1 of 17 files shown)


controller/internal/routes/current_identity_router.go
```diff
@@ -184,6 +203,8 @@ func (r *CurrentIdentityRouter) removeMfa(ae *env.AppEnv, rc *response.RequestCo
 		return
 	}
 
+	ae.Handlers.PostureResponse.SetMfaPostureForIdentity(rc.Identity.Id, false)
```
andrewpmartinez, line 206: On removing MFA from an identity this will set the posture data for mfa to "not passing" for all API Sessions.

dovholuknf (approved):

PR #602, opened 2021-03-23
check if enabled before trying to clean up

andrewpmartinez:


controller/server/controller.go
```diff
@@ -351,45 +348,30 @@ func (c *Controller) Run() {
 	}()
 }
```
plorenz, line 350: This is used in ziti-controller, assuming that's being updated

plorenz (approved):

controller/server/controller.go
andrewpmartinez: It is. I have a PR sitting waiting for this to go through

andrewpmartinez: Merged now.

PR #603, opened 2021-03-24
Host.v1 support

plorenz:


tunnel/intercept/hosting.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func (self *baseHostingContext) dialAddress(options map[string]interface{}, protocol string, address string) (net.Conn, error) {
+	var sourceIp string
+	if val, ok := options[SourceIpKey]; ok {
+		sourceIp = val.(string)
+	}
+
+	if sourceIp != "" {
+		dialer := net.Dialer{LocalAddr: &net.TCPAddr{IP: net.ParseIP(sourceIp)}}
```
scareything, line 93: could you set Timeout here?

plorenz: I missed that. Fixing.

PR #618, opened 2021-04-01
Support for intercept.v1 config type

plorenz:


tunnel/intercept/svcpoll.go
```diff
@@ -234,3 +257,104 @@ func (self *ServiceListener) host(svc *entities.Service) {
...
+		result := strings.ReplaceAll(sourceIp, sourceIpTemplKey, sourceAddrIp)
+		result = strings.ReplaceAll(result, sourcePortTemplKey, sourceAddrPort)
+		result = strings.ReplaceAll(result, destIpTemplKey, destAddrIp)
+		result = strings.ReplaceAll(result, destPortTemplKey, destAddrPort)
+		return result
+	}
+
+	return nil
+}
+
```
plorenz, line 311: @scareything note I changed this from tags to appData. Let me know if that's ok or if we want to keep it tags and just have it use appData under the hood

tunnel/intercept/interceptor.go
```diff
@@ -90,86 +88,21 @@ func addrBits(ip net.IP) int {
...
-	}
-	return &s, nil
-}
-
-func (m LUT) GetByName(serviceName string) []InterceptedService {
-	var r []InterceptedService
-	for _, service := range m {
-		if serviceName == service.Name {
-			r = append(r, service)
+func GetInterceptAddresses(service *entities.Service, protocol string, resolver dns.Resolver) ([]*InterceptAddress, error) {
```
scareything, line 91: Should protocol be populated with intercept.v1 details?
What about client.v1 configuration?

tunnel/intercept/source_ip_test.go
```diff
@@ -0,0 +1,69 @@
...
+		req.NoError(poller.configureSourceAddrProvider(svc))
+		req.Equal(expected, svc.GetSourceAddr(sourceAddr, destAddr))
+	}
+
+	testMatch("$src_ip:$src_port", "1.2.3.4:5432")
+	testMatch("$src_ip:$dst_port", "1.2.3.4:80")
+	testMatch("$dst_ip:$src_port", "5.6.7.8:5432")
+	testMatch("$dst_ip:$dst_port", "5.6.7.8:80")
+	testMatch("$tunneler_id.name", "foo.bar")
+	testMatch("$tunneler_id.tag[srcip]", "123.456.789.10:5555")
```
scareything, line 67: I guess you can populate the tags with any value you like, but fyi I would expect `$tunneler_id.tag[srcip]` to resolve to just an IP; if a port is also desired I would add it to the `sourceAddr` template.

tunnel/intercept/svcpoll.go
```diff
@@ -101,21 +114,42 @@ func (self *ServiceListener) HandleServicesChange(eventType ziti.ServiceEventTyp
...
-			pfxlog.Logger().Debugf("no service config of type %v for service %v", entities.ClientConfigV1, svc.Name)
-		} else if err != nil {
-			pfxlog.Logger().WithError(err).Errorf("error decoding service config of type %v for service %v", entities.ClientConfigV1, svc.Name)
+		if err != nil {
+			logrus.WithError(err).Errorf("error decoding service config of type %v for service %v", entities.InterceptV1, svc.Name)
+		}
+
+		if !found {
+			clientConfig := &entities.ServiceConfig{}
+			found, err = svc.GetConfigOfType(entities.ClientConfigV1, clientConfig)
```
scareything, line 135: Ahh. Disregard my previous question about client.v1. (I'm still looking for the place where intercept protocol is handled though)

tunnel/tunnel.go
```diff
@@ -17,21 +17,56 @@
...
+		return
+	}
+
+	if err := provider.TunnelService(service, clientConn, halfClose, appInfoJson); err != nil {
+		log.WithError(err).WithField("service", service.GetName()).Error("tunnel failed")
 		_ = clientConn.Close()
 	}
 }
 
+func GetIpAndPort(ipPort string) (string, string) {
```
scareything, line 52: I would think this wants to handle strings that only contain an IP.

tunnel/intercept/svcpoll.go
plorenz: tproxy_linux lines 140 and 149

tunnel/tunnel.go
plorenz: It's only being called with the results of net.Addr.String(), which AFAIK will always have a host port. Maybe should convert it to take net.Addr directly...

plorenz: Updated it to take net.Addr directly and try type casting first

tunnel/intercept/source_ip_test.go
plorenz: I figured it should work either way. It's probably not a common case. But it's just a template, so it's up the admins to settle on patterns that make sense.

scareything (approved):

PR #620, opened 2021-04-05
Per service identity hosting values

plorenz:


specs/swagger.yml
```diff
@@ -4566,6 +4566,10 @@ definitions:
         $ref: '#/definitions/terminatorPrecedence'
       defaultHostingCost:
         $ref: '#/definitions/terminatorCost'
+      serviceHostingPrecedences:
```
andrewpmartinez, line 4569: Do these need to be in the `required` block? Null values would mean these are trimmed from JSON response. You can either add these to the `required` section and then `null` + property name will be included. Otherwise, just ensure an empty map is always supplied would mean the property + `{}` would be supplied.

andrewpmartinez (approved):

specs/swagger.yml
plorenz: I'm ok having them dropped from the response when they are empty. But I'm happy to add them to required if you have a stylistic preference.

plorenz: Added them to required

PR #621, opened 2021-04-06
Add app data to identity and edge router. Add isSystem to ERP

plorenz:


specs/swagger.yml
```diff
@@ -4570,6 +4570,9 @@ definitions:
         $ref: '#/definitions/terminatorPrecedenceMap'
       serviceHostingCosts:
         $ref: '#/definitions/terminatorCostMap'
+      appData:
```
andrewpmartinez, line 4573: Should `appData` be in the `required` block?

- Not in required, zero values (nil) cause the field to be trimmed from the json reponse
- In required, zero values output as `null`

andrewpmartinez (approved):

PR #631, opened 2021-04-12
host.v1 configuration type schema changes

scareything:
* Rename host.v1 `dialIntercepted*` properties to `forward*`
* Add `allowed*` properties to whitelist destinations that are dialed via `forward*`
* Add `allowedSourceAddresses`, which serves as a whitelist for source IPs/CIDRs and informs the hosting tunneler of the local routes to establish when hosting a service

Caveats:
* Any existing host.v1 configurations will become invalid.
* This PR attempts to clean up local routes that were established for `allowedSourceAddresses` via `HostingContext.OnClose`; but I've found that this rarely works when called due to the process exiting. This may be reason to consider a unified exit handler strategy as a separate PR.


tunnel/entities/service.go
```diff
@@ -167,26 +181,70 @@ func (self *HostV2Terminator) getValue(options map[string]interface{}, key strin
...
 func (self *HostV2Terminator) GetPort(options map[string]interface{}) (string, error) {
-	if self.DialInterceptedPort {
-		return self.getValue(options, tunnel.DestinationPortKey)
+	if self.ForwardPort {
+		portStr, err := self.getValue(options, tunnel.DestinationPortKey)
+		if err != nil {
+			return portStr, err
+		}
+		port, err := strconv.Atoi(portStr)
+		for _, portRange := range self.AllowedPortRanges {
```
plorenz, line 225: Should this be `&&`?

scareything: yes, thanks!

plorenz:
Since we're making backward incompatible changes to host.v1, can we make the listenOptions connectTimeout be the same as that in host.v2?

plorenz (approved):

PR #632, opened 2021-04-12
adds the ability for router to extend their enrollment

andrewpmartinez:
- routers can now opt to extend their enrollment at any time with a
  previous or new key
- adds tests for router enrollment
- adds tests for router enrollment extension


plorenz (commented):
Would we gain anything from allowing routers to have multiple fingerprints, so that certs could overlap? Or would that just add complexity for no additional safety?

andrewpmartinez:
> Would we gain anything from allowing routers to have multiple fingerprints, so that certs could overlap? Or would that just add complexity for no additional safety?

I believe it wouldn't add value for the complexity it would add. 

The connection to the control channel (using the client cert) and all inbound from clients (using the server cert) only use the certs for the initial handshake and then they move to symmetric keys (a-la standard TLS, this is also true for links). Only new connections care about the certs. Additionally, the controller only allows 1 connection per router, so having multiple certs doesn't allow a smooth transition between newly negotiated connections.

The only reason to allow multiple certs IMO would be if the controller actually disconnected the old router connection - which is specifically avoided so that routing table/link churn doesn't happen. The security and identity have already been strongly validated, there is no reason to disconnect them. It is analogous to how a human can log in, change their password, and not be logged out; identity verification retained its continuity.

One reason for multiple certs is the corner case where the router/controller/connection between them fails exactly at the right time to cause the new certs to take hold in the controller, but the router doesn't receive the new certs. I plan on adding "re-enrollment" specifically to recover router's certs that were lost/expired for this type of scenario and hence why this was renamed "enrollment extension".

plorenz (approved):

PR #635, opened 2021-04-13
add policyType to posture query sets

andrewpmartinez:
- also fixes policy ids being set as the same value due to pointer in
  for loop reference

### DIFF
--- a/specs/swagger.yml
+++ b/specs/swagger.yml
@@ -3,7 +3,7 @@ swagger: '2.0'
 info:
   version: 0.16.3
   title: Ziti Edge
-  contact: {}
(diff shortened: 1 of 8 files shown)


specs/swagger.yml
```diff
@@ -3831,7 +3831,7 @@ responses:
       $ref: '#/definitions/currentAPISessionDetailEnvelope'
     examples:
       default:
-        meta: {}
+        meta: { }
```
andrewpmartinez, line 3834: looks like we have dueling formatting of yml.

plorenz (approved):

PR #656, opened 2021-05-04
splits API specs into composable modules

andrewpmartinez:
This should have no functional changes other than `/`  routing to `/edge/client/v1` and the introduction of `/edge/management/v1` both hosted on the same server. Separation of the interface/port of the APIs is part of the second branch/PR.

- splits Open API 2.0 (swagger) into client/management APIs
- uses go-swaggers flatten capabilities
- uses newer version of go-swagger (v0.27.0)
- introduces /edge/client/v1, /edge/management/v1
- non-prefixed URLs default to /edge/client/v1
- splits some shared management/client API handlers
- renames some REST models to "client" instead of "limited"
- passes all apitests, fails dataflow upd test (main fails this as well)
- updates generate rest script for powershell
- todo: test generate rest script for bash
- todo: use xweb (different branch)


plorenz:
dataflow tests pass for me on your branch. Guessing there's a windows issue with the test that's failing for you.

plorenz (approved):

PR #671, opened 2021-05-25
fixes deadlocks when a router disconnects mid sync or update

andrewpmartinez:
- range() of rtx'ers no longer locks, snapshots then loops
- rtx.Send() now checks for stopped/disconnected
- rtx.Send() now respects closeNotify

### DIFF
--- a/controller/env/broker.go
+++ b/controller/env/broker.go
@@ -75,14 +75,7 @@ func NewBroker(ae *AppEnv, synchronizer RouterSyncStrategy) *Broker {
 
 func (broker *Broker) RouterConnected(router *network.Router) {
 	go func() {
-
 		log := pfxlog.Logger().WithField("routerId", router.Id).WithField("routerName", router.Name).WithField("routerFingerprint", router.Fingerprint)
-
-		//check connection status, if already connected, ignore as it will be disconnected shortly
-		if broker.ae.IsEdgeRouterOnline(router.Id) {
-			log.Errorf("duplicate router connection detected [id: %s], ignoring", router.Id)
-			return
-		}
 
 		if router.Fingerprint == nil {
 			log.Errorf("router without fingerprints connecting [id: %s], ignoring", router.Id)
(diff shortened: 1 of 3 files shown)


controller/sync_strats/rtx.go
```diff
@@ -73,15 +73,15 @@ func (rtx *RouterSender) Start() {
 func (rtx *RouterSender) Stop() {
 	if rtx.stopping.CompareAndSwap(false, true) {
 		go func() {
```
plorenz, line 75: Usual pattern is to just close the `closeNotify` channel. Also means you shouldn't have to do it in a separate goroutine.

controller/env/broker.go
```diff
@@ -75,15 +75,8 @@ func NewBroker(ae *AppEnv, synchronizer RouterSyncStrategy) *Broker {
 
 func (broker *Broker) RouterConnected(router *network.Router) {
 	go func() {
-
 		log := pfxlog.Logger().WithField("routerId", router.Id).WithField("routerName", router.Name).WithField("routerFingerprint", router.Fingerprint)
```
plorenz, line 79: Will this allow existing connections to replaced? If so, is that desirable? If so, should we close the old connection?

andrewpmartinez: It can't be triggered w/ an existing connection based on how the fabric code reads and behaves.

I created a router with the same identity but two different sets of ports they listen on. The fabric refused to allow the second instance to run. Attempting to use the same configuration on the same machine causes port usage errors.

controller/sync_strats/rtx.go
andrewpmartinez: Changed

controller/env/broker.go
plorenz: Ah, gotcha. Makes sense to remove if it's redundant

plorenz (approved):

PR #676, opened 2021-06-01
fixes rtx double stop panic

andrewpmartinez:
- stopping status was being reset causing subsequent stops to panic
- remove stopping atomic bool
- change router disconnect logging from debug to info

### DIFF
--- a/controller/sync_strats/rtx.go
+++ b/controller/sync_strats/rtx.go
@@ -43,15 +43,19 @@ type RouterSender struct {
 }
 
 func newRouterSender(edgeRouter *model.EdgeRouter, router *network.Router, sendBufferSize int) *RouterSender {
-	return &RouterSender{
+	rtx := &RouterSender{
 		Id:          eid.New(),
 		EdgeRouter:  edgeRouter,
 		Router:      router,
 		send:        make(chan *channel2.Message, sendBufferSize),
-		closeNotify: nil,
-		running:     concurrenz.AtomicBoolean(0),
+		closeNotify: make(chan struct{}, 0),
+		running:     concurrenz.AtomicBoolean(1),
 		RouterState: env.NewLockingRouterStatus(),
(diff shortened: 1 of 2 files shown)


controller/sync_strats/rtx.go
```diff
@@ -66,22 +64,26 @@ func (rtx *RouterSender) GetState() env.RouterStateValues {
 
 func (rtx *RouterSender) Start() {
 	if rtx.running.CompareAndSwap(false, true) {
```
plorenz, line 66: I'd leave closeNotify init above. I don't see a reason to create it just-in-time or nil it out afterwards. Maybe I'm missing something, though?

andrewpmartinez: It was to allow the `RouterSender` to be re-stoppable and re-startable. I can make it not that as that functionality isn't used in the final version.

plorenz: I'd wait until we're using it to add that, just in case we never do :)

andrewpmartinez: Removed.

controller/sync_strats/rtx.go
```diff
@@ -38,22 +38,24 @@ type RouterSender struct {
...
-	return &RouterSender{
+	rtx := &RouterSender{
 		Id:          eid.New(),
 		EdgeRouter:  edgeRouter,
 		Router:      router,
 		send:        make(chan *channel2.Message, sendBufferSize),
 		closeNotify: make(chan struct{}, 0),
-		running:     concurrenz.AtomicBoolean(0),
-		stopping:    concurrenz.AtomicBoolean(0),
+		running:     concurrenz.AtomicBoolean(1),
```
andrewpmartinez, line 52: Defaults to running now

plorenz (approved):

PR #684, opened 2021-06-09
Delete expired api sessions in batches. Remove some unused methods

plorenz:


controller/model/base_handler.go
```diff
@@ -347,6 +347,17 @@ func (handler *baseHandler) deleteEntity(id string) error {
 	})
 }
 
+func (handler *baseHandler) deleteEntityBatch(ids []string) error {
```
andrewpmartinez, line 350: When we talked about this earlier I assumed by "batch" you meant using `db.Batch`, but what you meant was deleting multiple in the same db lock.

andrewpmartinez (approved):

controller/model/base_handler.go
plorenz: Yeah, since this isn't a user interactive process, I don't think we need to use Batch

Issue #670, opened 2021-05-25
ziti-tunnel leftover loopback IP assignments


qrkourier:
`ziti-tunnel` 0.20.2 leaves behind IP assignments to the loopback device e.g. `lo` for each intercept when it exits after a ^C interrupt signal .

```
$ ziti-tunnel tproxy --dnsSvcIpRange=100.64.192.0/18
...
^C

$ ip address show lo
...
    inet 100.64.192.1/32 scope host lo
       valid_lft forever preferred_lft forever
...
```

qrkourier:
I'm able to reproduce this issue in 0.20.9.

qrkourier:
@plorenz I reproduced this issue with v0.22.1 after finding it was necessary to manually clean up the leftover loopback assignments.

```bash
 /usr/bin/zsh:99> sudo ip addr del 100.64.64.1/32 dev lo
 /usr/bin/zsh:99> sudo ip addr del 100.64.64.2/32 dev lo
 /usr/bin/zsh:99> sudo ip addr del 100.64.64.3/32 dev lo
 /usr/bin/zsh:99> sudo ip addr del 100.64.64.4/32 dev lo
 /usr/bin/zsh:99> sudo ip addr del 100.64.64.5/32 dev lo
 /usr/bin/zsh:99> sudo ip addr del 100.64.64.6/32 dev lo
 /usr/bin/zsh:99> sudo ip addr del 100.64.64.7/32 dev lo
 /usr/bin/zsh:99> sudo ip addr del 100.64.64.8/32 dev lo
 /usr/bin/zsh:99> sudo ip addr del 100.64.64.9/32 dev lo
 /usr/bin/zsh:99> sudo ip addr del 100.64.64.10/32 dev lo
```

plorenz:
Reopening based on comment from @qrkourier

qrkourier:
Let me know if this is one of those things that "shouldn't be possible" :wink: 

If that's the case, then I would suspect some kind of ungraceful exit for `ziti-tunnel` and I could spend more time reproducing it with your guidance to look for particular conditions or patterns.

qrkourier:
I verified the `NF-INTERCEPT` subchain had been cleaned up, and so just the loopback assignments were still in effect.

qrkourier:
@plorenz I can still reproduce this is in 0.22.11.

scareything:
[...]

PR #691, opened 2021-06-15
Fix address remove from loopback on shutdown. Fixes GH-670

plorenz:


tunnel/intercept/tproxy/tproxy_linux.go
```diff
@@ -422,6 +419,7 @@ func (self *tProxy) addInterceptAddr(interceptAddr *intercept.InterceptAddress,
 		return fmt.Errorf("failed to add local route: %v", err)
 	}
 	tracker.AddAddress(ipNet.String())
```
plorenz, line 421: At some point we lost track of the addresses entirely

tunnel/intercept/tproxy/tproxy_linux.go
```diff
@@ -458,32 +456,36 @@ func (self *tProxy) addInterceptAddr(interceptAddr *intercept.InterceptAddress,
...
 			err = self.interceptor.ipt.Delete(filterTable, dstChain, addr.AcceptSpec...)
 			if err != nil {
-				return fmt.Errorf("failed to remove iptables rule for service %s: %v", self.service.Name, err)
+				errors = append(errors, err)
+				logrus.Errorf("failed to remove iptables rule for service %s: %v", self.service.Name, err)
 			}
 		}
-		ipn := addr.IpNet()
-		routes[ipn.String()] = ipn
-	}
```
plorenz, line 476: This was doing addresses uniquely, but we had the same address twice, once for UDP and once for TCP. If you only remove once, the address tracker returns false the first time because it knows it's used twice.

scareything:
It looks like this addresses the (relatively recent?) issue of intercept routes outliving the tunneler.

We also add addresses to `lo` for source IP spoofing with hosted services; and I think those can be left behind due to raciness in the exit handlers / hosting context close handlers.

scareything (approved):

scareything:
One of the advantages of the deprecated tun mode was that any routes that were assigned to the tun interface were automatically deleted when the tun interface was deleted (which the kernel did for us when the process exited for any reason). We route through loopback when using TPROXY, so we need to clean the routes up ourselves.

I was thinking that a more robust/failsafe way to clean up routes might be to set a shortish (30s) `valid_lft` on the addresses that are assigned to `lo`, and refresh the addresses from a goroutine in the tunneler (perhaps the service poll loop). The kernel would automatically delete any routes that weren't refreshed in the `valid_lft` interval.

qrkourier:
@scareything I believe that would address the problem of left-behind assignments from the DNS intercept IP range, but it wouldn't be possible to safely assume that an IP-only intercept address that'd been assigned to loopback e.g. 10.10.10.10 was detritus.

scareything:
@qrkourier The idea is that `valid_lft` would be set on any/all addresses that the tunneler assigns to `lo`, including the "dns" CIDR and individual IPs (and source IPs too). If the tunneler is adding the address then we know that it's either an intercept address or an allowed source address for a hosted service.

plorenz:
> One of the advantages of the deprecated tun mode was that any routes that were assigned to the tun interface were automatically deleted when the tun interface was deleted (which the kernel did for us when the process exited for any reason). We route through loopback when using TPROXY, so we need to clean the routes up ourselves.
> 
> I was thinking that a more robust/failsafe way to clean up routes might be to set a shortish (30s) `valid_lft` on the addresses that are assigned to `lo`, and refresh the addresses from a goroutine in the tunneler (perhaps the service poll loop). The kernel would automatically delete any routes that weren't refreshed in the `valid_lft` interval.

Sounds like a good idea. I took a quick peek and I think it'll be more refactoring and netlink digging than I want to do right now, so I'm going to leave it for the next round of refactoring.

scareything:
Yeah, it's definitely an idea for the next round of changes - or if stale routes become a more prevalent issue. BTW I was also thinking that it wouldn't be the end of the world if we shelled out to `ip address add` / `ip route add` instead of wrangling with the poorly documented netlink api.

plorenz:
> Yeah, it's definitely an idea for the next round of changes - or if stale routes become a more prevalent issue. BTW I was also thinking that it wouldn't be the end of the world if we shelled out to `ip address add` / `ip route add` instead of wrangling with the poorly documented netlink api.

Seems ok to do, as long as it doesn't introduce issues with varying versions of linux having/not having different utils installed. I suspect `ip` is pretty universal, but not sure.

qrkourier:
I think we're good with `ip` for now. It seems to be the modern, standard interface to `netlink` in all major distributions, and is typically installed as a package named `iproute2`.

PR #693, opened 2021-06-17
adds API tests to verify requestedName is used during ca auto enroll

andrewpmartinez:


tests/enrollment_ca_auto_test.go
```diff
@@ -128,11 +132,12 @@ func Test_IdentityEnrollment(t *testing.T) {
 			})
 
 			t.Run("can enroll with a name", func(t *testing.T) {
-				t.Run("can enroll without a name and empty JSON object", func(t *testing.T) {
+				t.Run("can enroll with", func(t *testing.T) {
```
dovholuknf, line 135: can enroll with 'what'?

dovholuknf: nevermind i see it referenced below now :)

tests/enrollment_ca_auto_test.go
```diff
@@ -143,12 +148,35 @@ func Test_IdentityEnrollment(t *testing.T) {
...
 						SetHeader("content-type", "application/json").
-						SetBody(`{"name": "` + eid.New() + `"}`).
+						SetBody(`{"name": "` + name + `"}`).
 						Post("enroll?method=ca")
 
 					ctx.Req.NoError(err)
 
 					ctx.Req.Equal(http.StatusOK, resp.StatusCode())
+
+					t.Run("the name was given to the identity", func(t *testing.T) {
```
dovholuknf, line 158: nit: the name given to the identity or the name that was given to the identity

tests/enrollment_ca_auto_test.go
```diff
@@ -143,12 +148,35 @@ func Test_IdentityEnrollment(t *testing.T) {
...
+						ctx.Req.NoError(err)
+						standardJsonResponseTests(resp, http.StatusOK, t)
+
+						envelope := rest_model.ListIdentitiesEnvelope{}
+
+						err = envelope.UnmarshalBinary(resp.Body())
+						ctx.Req.NoError(err)
+						found := false
+						for _, identity := range envelope.Data {
+							if *identity.Name == name {
```
dovholuknf, line 172: what happens when the same cased name is provided?

dovholuknf (approved):

tests/enrollment_ca_auto_test.go
andrewpmartinez: This is testing the ability for `[requestedName]`, which is provided during enrollment, is passed through to the controller and applied when required by the name formatting. The value passed in is used opaquely - so whatever the casing is, it is used.

PR #695, opened 2021-06-17
Add health checks

plorenz:


specs/management.yml
```diff
@@ -4479,6 +4479,38 @@ paths:
...
+      description: Returns OK or error depending on controller health status
+      tags:
+      - Informational
+      summary: Returns OK or error depending on controller health status
+      operationId: healthCheck
+      responses:
+        "200":
+          description: Base empty response
+          schema:
+            $ref: '#/definitions/empty'
```
andrewpmartinez, line 4494: I doubt this returns an empty response. Needs to point to the schema of whatever is returned. Useful for generated clients and testing.

specs/source/management/health-check.yml
```diff
@@ -0,0 +1,15 @@
...
+    get:
+      summary: Returns OK or error depending on controller health status
+      description: Returns OK or error depending on controller health status
+      security: [ ]
+      tags:
+        - Informational
+      operationId: healthCheck
+      responses:
+        '200':
+          $ref: '../shared/standard-responses.yml#/responses/emptyResponse'
```
andrewpmartinez, line 13: See comment from the flattened `management.yml`

specs/management.yml
andrewpmartinez: Also the resulting `rest_model.*` structs should be used in `controller/internal/routes/health_check_router.go` to format the output.

andrewpmartinez (changes requested):
See comments

controller/internal/routes/health_check_router.go
```diff
@@ -0,0 +1,141 @@
...
+func NewHealthCheckRouter() *HealthCheckRouter {
+	return &HealthCheckRouter{
+		BasePath:      "/health-check",
+		healthChecker: health.New(),
+	}
+}
+
+func (r *HealthCheckRouter) Register(ae *env.AppEnv) {
+	ae.ManagementApi.InformationalHealthCheckHandler = informational.HealthCheckHandlerFunc(func(params informational.HealthCheckParams) middleware.Responder {
+		return ae.IsAllowed(r.HealthCheck, params.HTTPRequest, "", "", permissions.Always())
```
andrewpmartinez, line 56: I understand why this is exposed publicly. If we ever get to adding API Keys for m2m this should be changed.

specs/management.yml
plorenz: It's returning as empty of a body as possible (using responder.RespondWithEmptyOk()). If everything is ok, it returns status code 200. If there's a problem it returns 503 with the error output from the health checker. I had the error bits encoded incorrectly, that's fixed now.

andrewpmartinez: Are you basing this pattern off something, it seems weird. I would expect a 200 ok to return a payload of 0 or more issues or the actual stats themselves.

Some frameworks that are generated from specs treat 4xx and 5xx responses as exceptions that must be caught. Seems weird to get an exception to catch it then look at what the payload is.

specs/management.yml
```diff
@@ -4479,6 +4479,48 @@ paths:
...
+                  field: (root)
+                  message: '(root): fooField3 is required'
+                  type: required
+                  value:
+                    fooField: abc
+                    fooField2: def
+                causeMessage: schema validation failed
+                code: COULD_NOT_VALIDATE
+                message: The supplied request contains an invalid document
+                requestId: ac6766d6-3a09-44b3-8d8a-1b541d97fdd9
```
andrewpmartinez, line 4520: I assume, based on your above comments, this example doesn't make sense.

specs/management.yml
```diff
@@ -4479,6 +4479,48 @@ paths:
...
+      tags:
+      - Informational
+      summary: Returns OK or error depending on controller health status
+      operationId: healthCheck
+      responses:
+        "200":
+          description: Base empty response
+          schema:
+            $ref: '#/definitions/empty'
+        "503":
```
andrewpmartinez, line 4495: I would expect a 503 to be returned only if the controller couldn't properly respond for some reason. If it is responding with the data requested, a 503 doesn't make sense.

specs/management.yml
plorenz: The health check library comes with an http serving library, which is used in the router. I didn't want to introduce a third potential endpoint for the controller so I added it to the management API. The built in API return 200 on healthy and 503 on unhealthy. It does output full check data on both success and fail. We only have one check right now so I didn't want to spend the time. I figured we can add more detailed output if add additional checks.

andrewpmartinez: 4xx and 5xx errors are meant to convey that the request could not be completed for some reason. 4xx for client correctable issues, 5xx for server-side issues.

Using 503 is not correct and in my opinion, would be unexpected.

If we only have 1 check  and the only viable output response is "good" or "bad", I suggest this response payload:

```
{
   "healthy": true/false
}
```

controller/internal/routes/health_check_router.go
```diff
@@ -0,0 +1,145 @@
...
+	rc.RespondWithApiError(err)
+}
+
+type boltPinger struct {
+	appEnv      *env.AppEnv
+	openReadTxs metrics.Gauge
+	running     concurrenz.AtomicBoolean
+}
+
+func (self *boltPinger) PingContext(ctx context.Context) error {
```
andrewpmartinez, line 113: Is anything using this? I can't find a reference to it. On first look, it seems like it doesn't belong at the REST level.

specs/management.yml
andrewpmartinez: I pulled the branch to see what the health check is doing. Seems pretty trivial to model their results generically.


```
responses:
  listHealthChecks:
    description: A list of controler health checks
    schema:
      $ref: '#/definitions/listHealthCheckEnvelope'

definitions:
   listHealthCheckEnvelope:
    type: object
    required:
      - meta
      - data
    properties:
      meta:
        $ref: '../shared/standard-responses.yml#/definitions/meta'
      data:
        $ref: '#/definitions/healthCheckResultList'
   healthCheckResultList:
      type: array
      items:
         $ref: '#/definitions/healCheckResultDetail'
   healthCheckResultDetail:
     type: object
     properties:
       message:
         type: string
       details:
          type: object
          additionalProperties: {}
        error:
          type: string
        timestamp:
          type: string
          format: date-time
       duration:
          type: string
          format: date-time
       contigousFailures:
          type: integer
       timeOfFirstFailure:
          type: string
          format: date-time
```

controller/internal/routes/health_check_router.go
plorenz: It's in Register

plorenz:
Superceded by fabric PR

PR #713, opened 2021-07-14
Don't remove ipt chains unless all services are removed. Don't use dn…

plorenz:
…s resolved address if it's a tunneler assigned addr. Fixes GH-712

### DIFF
--- a/tunnel/dns/server.go
+++ b/tunnel/dns/server.go
@@ -23,6 +23,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"net"
 	"net/url"
+	"os/exec"
 	"strings"
 	"sync"
 	"time"
@@ -36,7 +37,23 @@ type resolver struct {
 	namesMtx sync.Mutex
 }
 
+func flushDnsCaches() {
+	bin, err := exec.LookPath("systemd-resolve")
+	if err != nil {
+		logrus.WithError(err).Warn("unable to find systemd-resolve in path, consider adding a dns flush to your restart process")
+		return
+	}
+
+	cmd := exec.Command(bin, "--flush-caches")
+	if err = cmd.Run(); err != nil {
+		logrus.WithError(err).Warn("unable to flush dns caches, consider adding a dns flush to your restart process")
+	} else {
+		logrus.Info("dns caches flushed")
+	}
+}
+
 func NewResolver(config string) Resolver {
+	flushDnsCaches()
 	if config == "" {
 		return nil
 	}
--- a/tunnel/intercept/interceptor.go
+++ b/tunnel/intercept/interceptor.go
@@ -17,6 +17,7 @@
 package intercept
 
 import (
+	"fmt"
 	"github.com/openziti/edge/tunnel"
 	"github.com/openziti/edge/tunnel/dns"
 	"github.com/openziti/edge/tunnel/entities"
@@ -70,6 +71,11 @@ func (addr *InterceptAddress) HighPort() uint16 {
 
 func (addr *InterceptAddress) Contains(ip net.IP, port uint16) bool {
 	return addr.cidr.Contains(ip) && port >= addr.lowPort && port <= addr.highPort
+}
+
+func (addr *InterceptAddress) String() string {
+	return fmt.Sprintf("cidr: %v, cidrAddr: %p, lowPort: %v, highPort: %v, protocol: %v, tproxySpec: %v, acceptSpec: %v",
+		addr.cidr, addr.cidr, addr.lowPort, addr.highPort, addr.protocol, addr.TproxySpec, addr.AcceptSpec)
 }
 
 func GetInterceptAddresses(service *entities.Service, protocol string, resolver dns.Resolver) ([]*InterceptAddress, error) {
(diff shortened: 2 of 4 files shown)


tunnel/intercept/iputils.go
```diff
@@ -66,10 +66,14 @@ func getInterceptIP(hostname string, resolver dns.Resolver) (net.IP, *net.IPNet,
 	if err == nil {
 		if len(addrs) > 0 {
 			ip := addrs[0].To4()
```
plorenz, line 68: I hit a case where two services where sharing an IP because I did the following

1. intercepted service A with hostname foo. It got address 100.64.0.3
2. Stopped intercepting service A
3. There may have been a tunneler restart in here, not sure
4. Intercepted service B with hostname bar. It got address 100.64.0.3
5. Intercepted service A with hostname foo. Foo still resolved to 100.64.0.3, so it also got the same IP address

This checks to make sure that a DNS resolve isn't coming out of our IP addr range, assuming that if it does we've got a state DNS entry. If that's detected we'll assign a new IP addr to avoid conflicts.

tunnel/intercept/tproxy/tproxy_linux.go
```diff
@@ -106,6 +106,7 @@ func (self *interceptor) Stop() {
 		proxy.Stop(alwaysRemoveAddressTracker{})
```
plorenz, line 106: We were removing the NF-INTERCEPT chain whenever we removed any service, which caused all service intercepts to get cleaned up.

tunnel/utils/ipcalc.go
```diff
@@ -17,19 +17,29 @@
...
+func IsLocallyAssigned(addr, lower, upper net.IP) bool {
+	return bytes.Compare(addr.To16(), lower.To16()) >= 0 && bytes.Compare(addr.To16(), upper.To16()) <= 0
+}
+
 // return the next available IP address in the range of provided IPs
 func NextIP(lower, upper net.IP) (net.IP, error) {
 	usedAddrs, err := AllInterfaceAddrs()
 	if err != nil {
 		return nil, err
 	}
```
plorenz, line 36: We were always returning `lower` from this call, since lower is just a byte array. Additionally we were mutating lower as we were operating. This meant that each new service would end up change the value for all other intercept addresses, which would cause cleanup to fail.

andrewpmartinez:
I'm hesitant to approve this without someone w/ more tunneler experience to look at it.

tunnel/dns/server.go
```diff
@@ -36,7 +37,23 @@ type resolver struct {
 	namesMtx sync.Mutex
 }
 
+func flushDnsCaches() {
+	bin, err := exec.LookPath("systemd-resolve")
```
dovholuknf, line 41: does this 'force' us to use debian (ubuntu?) images?  not a problem for us or users using our appliance but for customers "doing it on their own" they should at least know this but i don't know how to communicate that to the end-user effectively... :/

plorenz: It won't break anything, it'll just spit out an error message if can't run it. Should I enhance the error message to say something like: `'error flushing dns, consider adding a dns-flush to your restart script'`?

tunnel/intercept/interceptor.go
```diff
@@ -72,6 +73,11 @@ func (addr *InterceptAddress) Contains(ip net.IP, port uint16) bool {
 	return addr.cidr.Contains(ip) && port >= addr.lowPort && port <= addr.highPort
 }
 
+func (addr *InterceptAddress) String() string {
+	return fmt.Sprintf("cidr: %v, cidrAddr: %p, lowPort: %v, highPort: %v, protocol: %v, tproxySpec: %v, acceptSpec: %v",
```
dovholuknf, line 77: should this have a nil guard?

tunnel/intercept/tproxy/tproxy_linux.go
```diff
@@ -122,10 +128,20 @@ func (self *interceptor) StopIntercepting(serviceName string, tracker intercept.
...
 		proxy.Stop(tracker)
 		self.serviceProxies.Remove(serviceName)
+		self.cleanupChains()
 	}
 	return nil
 }
 
+func (self *interceptor) cleanupChains() {
+	if self.serviceProxies.IsEmpty() {
+		deleteIptablesChain(self.ipt, mangleTable, "PREROUTING", dstChain)
```
dovholuknf, line 138: is there any sort of reference counting done (it's not clear in just this review) that some other service doesn't use the same hostname? it's perfectly fine for two services from two identities to use "my-intercept" as long as both identities were to use different ports. same is true for one identity with multiple services using "my-intercept". you probably do not want to delete the rules unless ALL the services using that host are gone

tunnel/intercept/interceptor.go
plorenz: For if the addr is null? Shouldn't ever called on a nil addr. It's only used for debugging what's in the list of addresses and the list of addresses should never contain nils

dovholuknf (approved):

tunnel/intercept/tproxy/tproxy_linux.go
plorenz: This is cleaning up the chains that are used by all services, so we only remove them if we have zero services being intercepted. 
The other rules are per-port range, so we remove those as specific port ranges are removed. 
Allocated IP addresses are per DNS entry and those are reference counted using the address tracker.

PR #741, opened 2021-08-06
Add more logging context and tracing

plorenz:
… path selection/establishment


controller/env/appenv.go
```diff
@@ -232,10 +233,10 @@ func (ae *AppEnv) FillRequestContext(rc *response.RequestContext) error {
 	if rc.SessionToken != "" {
 		_, err := uuid.Parse(rc.SessionToken)
 		if err != nil {
-			logger.WithError(err).Debug("failed to parse session id")
+			logger.WithError(err).Debug("failed to parse session traceId")
 			rc.SessionToken = ""
 		} else {
-			logger.Tracef("authorizing request using session id '%v'", rc.SessionToken)
+			logger.Tracef("authorizing request using session traceId '%v'", rc.SessionToken)
```
andrewpmartinez, line 239: "traceId" ?

specs/management.yml
```diff
@@ -6023,6 +6023,101 @@ paths:
...
+        name: traceSpec
+        in: body
+        required: true
+        schema:
+          $ref: '#/definitions/traceSpec'
+      responses:
+        "200":
+          description: Returns the document that represents the trace state
+          schema:
+            $ref: '#/definitions/traceDetailEnvelope'
```
andrewpmartinez, line 6048: For the rest of the API a `PUT` to a document URL returns an `emptyResponse` from the shared responses.

specs/source/management/identities.yml
```diff
@@ -355,6 +355,35 @@ paths:
...
+      parameters:
+        - name: traceSpec
+          in: body
+          required: true
+          description: A traceSpec object
+          schema:
+            $ref: '#/definitions/traceSpec'
+      responses:
+        '200':
+          $ref: '#/responses/traceDetail'
```
andrewpmartinez, line 380: For the rest of the API a `PUT` to a document URL returns an `emptyResponse` from the shared responses.

specs/source/management/identities.yml
```diff
@@ -355,6 +355,35 @@ paths:
           $ref: '../shared/standard-responses.yml#/responses/notFoundResponse'
         '401':
           $ref: '../shared/standard-responses.yml#/responses/unauthorizedResponse'
+  identities-id-trace:
+    parameters:
+      - $ref: '../shared/parameters.yml#/id'
+    put:
```
andrewpmartinez, line 361: Is there a reason PUT was chosen over POST?

PUT is usually associated with "create with id" and "update with id". This feels like a POST to create a new data entry. The odd part is that you can't read the current details of the trace (i.e. no GET).

We have this same pattern with MFA supporting POST and GET.

andrewpmartinez: Mmmm I can wrangle my brain into PUT being okay. Since there is only one record and no id necessary.

specs/source/management/identities.yml
andrewpmartinez: I see you are returning a different object with a calculated until value.

controller/env/appenv.go
plorenz: Goland rename went rouge. Fixing. It got a bunch of other ids in comments that I need to fix

plorenz: fixed

specs/source/management/identities.yml
plorenz: It's kind of a weird middle case. I don't have a strong preference. If you like POST better, I can switch it. I thought of it more as attaching some data to identity, rather than a standalone entity, but you can see it either way.

specs/source/management/identities.yml
plorenz: Yes, also if you don't provide a traceId, it will generate one for you.

andrewpmartinez (approved):

PR #756, opened 2021-08-19
Add entity count events

plorenz:
Output looks like 

```
{
  "namespace": "edge.entityCounts",
  "timestamp": "2021-08-19T13:39:54.056181406-04:00",
  "counts": {
    "apiSessionCertificates": 0,
    "apiSessions": 9,
    "authenticators": 4,
    "cas": 0,
    "configTypes": 5,
    "configs": 2,
    "edgeRouterPolicies": 4,
    "enrollments": 0,
    "eventLogs": 0,
    "geoRegions": 17,
    "identities": 6,
    "identityTypes": 4,
    "mfas": 0,
    "postureCheckTypes": 5,
    "postureChecks": 0,
    "routers": 2,
    "serviceEdgeRouterPolicies": 2,
    "servicePolicies": 5,
    "services": 3,
    "sessions": 0
  },
  "error": ""
}
```

### DIFF
--- a/events/entity_counts.go
+++ b/events/entity_counts.go
@@ -1,0 +1,82 @@
+package events
+
+import (
+	"fmt"
+	"github.com/openziti/edge/controller/persistence"
+	"github.com/pkg/errors"
+	"go.etcd.io/bbolt"
+	"reflect"
+	"time"
+)
+
+const EntityCountEventNS = "edge.entityCounts"
+
+type EntityCountEvent struct {
+	Namespace string           `json:"namespace"`
+	Timestamp time.Time        `json:"timestamp"`
+	Counts    map[string]int64 `json:"counts"`
+	Error     string           `json:"error"`
+}
+
+func (event *EntityCountEvent) String() string {
+	return fmt.Sprintf("%v timestamp=%v counts=%+v err=%v",
+		event.Namespace, event.Timestamp, event.Counts, event.Error)
+}
+
+var entityCountEventGenerator func(interval time.Duration, handler EntityCountEventHandler)
+
+func generateEntityCountEvent(dbProvider persistence.DbProvider, stores *persistence.Stores, handler EntityCountEventHandler) {
+	event := &EntityCountEvent{
+		Namespace: EntityCountEventNS,
+		Timestamp: time.Now(),
+		Counts:    map[string]int64{},
+	}
+
+	for _, store := range stores.GetStoreList() {
+		err := dbProvider.GetDb().View(func(tx *bbolt.Tx) error {
+			_, count, err := store.QueryIds(tx, "true limit 1")
+			if err != nil {
+				return err
+			}
+			event.Counts[store.GetEntityType()] = count
+			return nil
+		})
+
+		if err != nil {
+			event.Error = err.Error()
+			break
+		}
+	}
+
+	handler.AcceptEntityCountEvent(event)
+}
+
+type EntityCountEventHandler interface {
+	AcceptEntityCountEvent(event *EntityCountEvent)
+}
+
+func registerEntityCountEventHandler(val interface{}, config map[interface{}]interface{}) error {
+	handler, ok := val.(EntityCountEventHandler)
+
+	if !ok {
+		return errors.Errorf("type %v doesn't implement github.com/openziti/edge/events/EntityCountEventHandler interface.", reflect.TypeOf(val))
+	}
+
+	interval := time.Minute * 5
+
+	if val, ok := config["interval"]; ok {
+		if strVal, ok := val.(string); ok {
+			var err error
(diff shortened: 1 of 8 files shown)


mikegorman-nf (approved):
Thanx for the quick work!

events/entity_counts.go
```diff
@@ -0,0 +1,97 @@
...
+	AcceptEntityCountEvent(event *EntityCountEvent)
+}
+
+var entityCountEventHandlerRegistry = cowslice.NewCowSlice(make([]EntityCountEventHandler, 0))
+
+func getEntityCountEventHandlers() []EntityCountEventHandler {
+	return entityCountEventHandlerRegistry.Value().([]EntityCountEventHandler)
+}
+
+func registerEntityCountEventHandler(val interface{}, config map[interface{}]interface{}) error {
```
andrewpmartinez, line 72: Why take in `val` then type assert it to `EntityCountEventHandler` instead of defining the function as `val EntityCountEventHandler` ?

plorenz: The registration function has to conform to 

`type registrationHandler func(handler interface{}, config map[interface{}]interface{}) error`

This generic type is used by all the event types, so it can't be typed to a specific one.

events/entity_counts.go
```diff
@@ -0,0 +1,97 @@
...
+	Counts    map[string]int64 `json:"counts"`
+	Error     string           `json:"error"`
+}
+
+func (event *EntityCountEvent) String() string {
+	return fmt.Sprintf("%v timestamp=%v counts=%+v err=%v",
+		event.Namespace, event.Timestamp, event.Counts, event.Error)
+}
+
+var entityCountEventGenerator func(interval time.Duration)
```
andrewpmartinez, line 27: I only see this declared here and set once in `Initi(...)`. Is this for tests?

andrewpmartinez: The only other reason I can see is so you can use the closure scope for the closeNotify chan.

/shrug

mguthrie88 (approved):

events/entity_counts.go
```diff
@@ -0,0 +1,97 @@
...
+			interval, err = time.ParseDuration(strVal)
+			if err != nil {
+				return errors.Wrapf(err, "invalid duration value for edge.entityCounts interval: '%v'", strVal)
+			}
+		} else {
+			return errors.Errorf("invalid type %v for edge.entityCounts interval configuration", reflect.TypeOf(val))
+		}
+	}
+
+	go entityCountEventGenerator(interval)
```
andrewpmartinez, line 93: This function calls `AddEntityCountEventHandler` which adds to `entityCountEventHandlerRegistry`. 

Then  `go entityCountEventGenerator`  calls `generateEntityCountEvent` which in turn emits events for all handlers by calling `getEntityCountEventHandlers` which comes from `entityCountEventHandlerRegistry. 

Doesn't this mean that as `registerEntityCountEventHandler` is called multiple times every configured interval will trigger every other handler?

events/entity_counts.go
plorenz: It's used in registerEntityCountEventHandler

andrewpmartinez (changes requested):
See my comments, want to make sure I'm not missing something.

events/entity_counts.go
plorenz: Not any more :) 

I had considered this, but with the typical configuration, it wouldn't be a problem since we'd only ever have a single handler. But it was bothering me as well, so I fixed it. This means I removed the Add/Remove handlers, so this is a bit different from the other event types. However, if someone is in-process, they don't need this is event, they can query for it directly, so I think that's fine.

events/entity_counts.go
plorenz: It's just a place to store the state until register gets called.

andrewpmartinez: I didn't see why you would use a module-level `var` declared function over a regularly defined function other than the close notify closure scope.

It would be just as easy to use the close notify as a package-level variable. The only reason I've really seen `var` declared module functions is to monkey patch during testing.

plorenz: it's not just the close notify, it's also holding on to the the provider and stores. I could either store them all in individual vars, or a struct, or a func, which has the benefit of being a single var and opaque.

andrewpmartinez: Thanks, no change is needed. 

Interesting pattern. I would have encapsulated that data into another struct that had its own `Run()` or like function. Would do the same to facilitate scope and testing w/o using module-level overrides.

events/entity_counts.go
andrewpmartinez: Nice, good update.

andrewpmartinez (approved):

PR #764, opened 2021-09-01
Fix time check on list tunnel. Add policy enforcer test. Add update s…

plorenz:
…ession test


controller/handler_edge_ctrl/list_tunnel_services.go
```diff
@@ -83,7 +81,7 @@ func (self *listTunnelServicesHandler) listServices(ctx *listTunnelServicesReque
 		lastUpdate = self.appEnv.StartupTime
 	}
```
plorenz, line 83: This would skip if the timestamps were the same, so we were always sending service lists.

andrewpmartinez (approved):

PR #767, opened 2021-09-03
Allow ERT to gracefully handle api session/session removal. Can happe…

plorenz:
…n if router is disconnected, but not down for an extended period

### DIFF
--- a/controller/handler_edge_ctrl/remove_tunnel_terminator.go
+++ b/controller/handler_edge_ctrl/remove_tunnel_terminator.go
@@ -61,11 +61,13 @@ func (self *removeTunnelTerminatorHandler) RemoveTerminator(ctx *RemoveTunnelTer
 		return
 	}
 
-	ctx.verifyTerminator(ctx.terminatorId, edge_common.TunnelBinding)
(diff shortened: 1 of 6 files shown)


controller/handler_edge_ctrl/remove_tunnel_terminator.go
```diff
@@ -61,12 +61,14 @@ func (self *removeTunnelTerminatorHandler) RemoveTerminator(ctx *RemoveTunnelTer
...
 	}
 
-	ctx.verifyTerminator(ctx.terminatorId, edge_common.TunnelBinding)
+	t := ctx.verifyTerminator(ctx.terminatorId, edge_common.TunnelBinding)
 	if ctx.err != nil {
 		self.returnError(ctx, ctx.err)
 		return
 	}
 
+	logger = logger.WithField("serviceId", t.Service)
```
andrewpmartinez, line 70: Making sure this is actually correct. `Service` = "serviced"

andrewpmartinez (approved):

controller/handler_edge_ctrl/remove_tunnel_terminator.go
plorenz: Yes, double checked.

PR #769, opened 2021-09-08
Fix panic when service not found

plorenz:


controller/handler_edge_ctrl/common_tunnel.go
```diff
@@ -150,6 +150,29 @@ func (self *baseTunnelRequestContext) ensureApiSession(configTypes []string) boo
...
+		self.service, err = self.handler.getAppEnv().Handlers.EdgeService.ReadByName(name)
+
+		if err != nil {
+			if boltz.IsErrNotFoundErr(err) {
+				err = InvalidServiceError{}
+			} else {
+				err = internalError(err)
+			}
+
+			logrus.
```
plorenz, line 165: We were logging self.session.Id and self.session.ServiceId and self.session wasn't set yet at this point.

andrewpmartinez (approved):

PR #780, opened 2021-09-13
adds EventFullyAuthenticated to the API session store

andrewpmartinez:
The previous implementation would send API Sessions to Edge Routers while partially authenticated. Due to state issues addressed separately in the c-SDK, these partial API Sessions are maintained if a channel connected to an edge router was established during reconnect. This would further cause problems with channel connection using the wrong API session vs what the rest of the SDK was using.

### DIFF
--- a/controller/persistence/api_session_store.go
+++ b/controller/persistence/api_session_store.go
@@ -127,7 +127,7 @@ func (store *apiSessionStoreImpl) Update(ctx boltz.MutateContext, entity boltz.E
 
 	if err == nil {
 		if apiSession, ok := entity.(*ApiSession); ok && apiSession != nil {
-			if checker.IsUpdated(FieldApiSessionMfaComplete) && apiSession.MfaComplete == true {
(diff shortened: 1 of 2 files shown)


plorenz (approved):

controller/persistence/api_session_store.go
```diff
@@ -106,6 +109,33 @@ type apiSessionStoreImpl struct {
...
+	}
+
+	return err
+}
+func (store *apiSessionStoreImpl) Update(ctx boltz.MutateContext, entity boltz.Entity, checker boltz.FieldChecker) error {
+	err := store.baseStore.Update(ctx, entity, checker)
+
+	if err == nil {
+		if apiSession, ok := entity.(*ApiSession); ok && apiSession != nil {
+			if (checker == nil || checker.IsUpdated(FieldApiSessionMfaComplete)) && apiSession.MfaComplete == true {
```
andrewpmartinez, line 130: I was not expecting if `checker == nil` we allow updating all fields, but that is what the lower `ProceedWithSet` function does.

PR #794, opened 2021-10-04
Dns wildcard intercept support

ekoby:


tunnel/dns/file.go
```diff
@@ -73,9 +82,7 @@ func (h *hostFile) AddHostname(hostname string, ip net.IP) error {
 	return nil
 }
 
-func (h *hostFile) RemoveHostname(s string) error {
-	h.mutex.Lock()
-	defer h.mutex.Unlock()
+func (h *hostFile) RemoveHostname(_ string) error {
```
andrewpmartinez, line 85: Does this ...need to exist? Are we satisfying an interface?

andrewpmartinez: Ah, I see it below. If there aren't other impls that need it it is questionable if the interface needs it.

However, no change needed imo.

tunnel/dns/server.go
```diff
@@ -98,7 +101,14 @@ func NewDnsServer(addr string) Resolver {
...
 
 	names := make(map[string]net.IP)
-	r := &resolver{server: s, names: names, namesMtx: sync.Mutex{}}
+	r := &resolver{
+		server:     s,
+		names:      names,
+		ips:        make(map[string]string),
+		namesMtx:   sync.Mutex{},
+		domains:    make(map[string]*domainEntry),
+		domainsMtx: sync.Mutex{},
```
andrewpmartinez, line 110: Why this instead of a concurrent map? I assume to avoid having to deal with `interfac{}`

PR #799, opened 2021-10-08
added geneve proxy xgress

dariuszSki:


router/xgress_geneve/listener.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+type listener struct{}
+
+func (self *listener) Listen(string, xgress.BindHandler) error {
+	go func() {
+		log := pfxlog.Logger()
+		// Open UDP socket to listen for Geneve Packets
+		conn, err := net.ListenPacket("udp", ":6081")
+		if err != nil {
```
plorenz, line 37: Once you've got the `WithError` you don't need it in the message anymore.

`log.WithError(err).Error("failed to open geneve interface - udp")`

dariuszSki: Good catch thanks!, updated it

router/xgress_geneve/listener.go
```diff
@@ -0,0 +1,108 @@
...
+			log.WithError(err).Errorf("failed to open geneve interface - udp")
+			// error but return gracefully
+			return
```
plorenz, line 40: Can remove the `else` from around the log statement

router/xgress_geneve/listener.go
```diff
@@ -0,0 +1,108 @@
...
+			log.Infof("geneve interface started successfully - udp: %s", conn.LocalAddr().String())
+		}
+		// Close it when done
+		defer conn.Close()
+		// Open a raw socket to send Modified Packets to Networking Stack
+		fd, err := syscall.Socket(syscall.AF_INET, syscall.SOCK_RAW, syscall.IPPROTO_RAW)
+		if err != nil {
+			log.WithError(err).Errorf("failed to open geneve interface - fd")
+			// error but return gracefully
+			return
```
plorenz, line 51: Can remove the `else` from around the log statement

dariuszSki: good point, will eliminate the else

PR #818, opened 2021-10-21
adds identity cert extension endpoint to client/man API

andrewpmartinez:
- now supports POST /current-identity/authenticators/{id}/extend for
  cert authenticators
- adds test for cert extension endpoint


plorenz:
Why not let an identity have multiple (or at least two) valid certs? Is it b/c it makes the implementation harder or because there are security concerns? Seems like you could more easily end up with a broken client if there's only one valid allowed at a time, if for example the client gets interrupted during the request.

ekoby:
we also need to add an ability to update 3rd party client certs

andrewpmartinez:
> Why not let an identity have multiple (or at least two) valid certs? Is it b/c it makes the implementation harder or because there are security concerns? Seems like you could more easily end up with a broken client if there's only one valid allowed at a time, if for example the client gets interrupted during the request.

The data model allows an identity to have any number of authenticators. However, creating authenticators has traditionally been limited to admins and is only possible in the Management API. Allowing identities to create an unbounded number of authenticators seems odd. Creating a limited number (i.e., 2 as you suggested) - only helps if no one ever creates additional authenticators for an identity. It would also break if someone did (i.e., creating then deleting with a limit of two would fail if a second had already been created). 

If the goal is to ensure that certificate updates aren't lost, I'd rather add a confirmation step with a delayed update. This would involve changing authenticators to have a pending new cert on them. The client would have to track/respect this state and provide a final confirmation to use the new cert.  It is more complex, and the fail case has more edge cases.

andrewpmartinez:
> we also need to add an ability to update 3rd party client certs

Sure do; that would be a different endpoint because there is no CSR involved.

plorenz (approved):

PR #831, opened 2021-11-09
Remove some unused types. Fix version map

plorenz:


controller/response/responder.go
```diff
@@ -151,8 +151,8 @@ func RespondWithApiError(w http.ResponseWriter, r *http.Request, requestId strin
 	data := &rest_model.APIErrorEnvelope{
 		Error: apierror.ToRestModel(apiError, requestId),
 		Meta: &rest_model.Meta{
```
plorenz, line 153: Should I fix the spelling of enrollment while I'm in here @andrewpmartinez ?

andrewpmartinez: Oh my. Yes - Nothing keys off it yet.

andrewpmartinez (approved):

PR #838, opened 2021-11-18
Initial pass at fabric rest api working

plorenz:


controller/internal/routes/base_router.go
```diff
@@ -237,7 +172,7 @@ func CreateWithResponder(rc *response.RequestContext, rsp response.Responder, li
 			return
 		}
 
-		if sve, ok := err.(*schema.ValidationErrors); ok {
+		if sve, ok := err.(*apierror2.ValidationErrors); ok {
```
andrewpmartinez, line 175: apierror2 looks like a refactor collission

andrewpmartinez:
Lots of edge dupe'd code here. Over time we should probably move to 1 copy and/or move bits to a shared lib.

andrewpmartinez (approved):

controller/internal/routes/base_router.go
plorenz: Went through and fixed these

PR #859, opened 2022-01-06
adds eventual eventer to controller

andrewpmartinez:
Isn't hooked up to replace the ApiSession/Session delete logic yet, but this would allow any event to be persisted to bbolt, raised at a later date, and then deleted.

Edit: Now is hooked up to API session/session delete

### DIFF
--- a/controller/model/testing.go
+++ b/controller/model/testing.go
@@ -120,10 +120,6 @@ func (ctx *TestContext) GetFingerprintGenerator() cert.FingerprintGenerator {
 }
 
 func NewTestContext(t *testing.T) *TestContext {
-	return NewTestContextWithDb(t, "")
-}
-
-func NewTestContextWithDb(t *testing.T, dbPath string) *TestContext {
 	context := &TestContext{
 		TestContext:     persistence.NewTestContext(t),
 		metricsRegistry: metrics.NewRegistry("test", nil),
@@ -131,8 +127,10 @@ func NewTestContextWithDb(t *testing.T, dbPath string) *TestContext {
 			closeNotify: make(chan struct{}),
 		},
 	}
-	context.InitWithDbFile(dbPath)
 	return context
+}
+func (ctx *TestContext) Init() {
+	ctx.InitWithDbFile("")
 }
 
 func (ctx *TestContext) InitWithDbFile(dbPath string) {
--- a/controller/persistence/api_session_store.go
+++ b/controller/persistence/api_session_store.go
@@ -17,7 +17,9 @@
 package persistence
 
 import (
+	"fmt"
 	"github.com/kataras/go-events"
+	"github.com/michaelquigley/pfxlog"
 	"github.com/openziti/edge/eid"
 	"github.com/openziti/foundation/storage/ast"
 	"github.com/openziti/foundation/storage/boltz"
@@ -34,7 +36,8 @@ const (
 	FieldApiSessionMfaRequired    = "mfaRequired"
 	FieldApiSessionLastActivityAt = "lastActivityAt"
 
-	EventFullyAuthenticated events.EventName = "FULLY_AUTHENTICATED"
+	EventFullyAuthenticated       events.EventName = "FULLY_AUTHENTICATED"
+	EventualEventApiSessionDelete                  = "ApiSessionDelete"
 )
 
 type ApiSession struct {
@@ -98,6 +101,8 @@ func newApiSessionStore(stores *stores) *apiSessionStoreImpl {
 	store := &apiSessionStoreImpl{
 		baseStore: newBaseStore(stores, EntityTypeApiSessions),
 	}
+
+	stores.EventualEventer.AddEventualListener(EventualEventApiSessionDelete, store.onEventualDelete)
 	store.InitImpl(store)
 	return store
 }
@@ -107,6 +112,40 @@ type apiSessionStoreImpl struct {
 
 	indexToken     boltz.ReadIndex
 	symbolIdentity boltz.EntitySymbol
+}
+
+func (store *apiSessionStoreImpl) onEventualDelete(name string, data []byte) {
+	var ids []string
+	err := store.stores.DbProvider.GetDb().View(func(tx *bbolt.Tx) error {
+		query := fmt.Sprintf(`%s = "%s"`, FieldSessionApiSession, string(data))
+		var err error
+		ids, _, err = store.stores.session.QueryIds(tx, query)
+		return err
+	})
+
+	if err != nil {
+		pfxlog.Logger().WithError(err).WithFields(map[string]interface{}{
+			"eventName":    name,
+			"apiSessionId": string(data),
+		}).Error("error querying for session associated to an api session during onEventualDelete")
+	}
+
+	for _, id := range ids {
+		_ = store.stores.DbProvider.GetDb().Update(func(tx *bbolt.Tx) error {
+			ctx := boltz.NewMutateContext(tx)
+			err := store.stores.session.DeleteById(ctx, id)
+
+			if err != nil {
+				pfxlog.Logger().WithError(err).WithFields(map[string]interface{}{
+					"eventName":    name,
+					"apiSessionId": string(data),
+					"sessionId":    id,
+				}).Error("error deleting for session associated to an api session during onEventualDelete")
+			}
+
+			return nil
+		})
+	}
 }
 
 func (store *apiSessionStoreImpl) Create(ctx boltz.MutateContext, entity boltz.Entity) error {
@@ -130,6 +169,20 @@ func (store *apiSessionStoreImpl) Update(ctx boltz.MutateContext, entity boltz.E
 			if (checker == nil || checker.IsUpdated(FieldApiSessionMfaComplete)) && apiSession.MfaComplete == true {
 				store.Emit(EventFullyAuthenticated, apiSession)
 			}
+		}
+	}
+
+	return err
+}
+
+func (store *apiSessionStoreImpl) DeleteById(ctx boltz.MutateContext, id string) error {
+	err := store.baseStore.DeleteById(ctx, id)
+
+	if err == nil {
+		if bboltEventualEventer, ok := store.baseStore.stores.EventualEventer.(*EventualEventerBbolt); ok {
+			bboltEventualEventer.AddEventualEventWithCtx(ctx, EventualEventApiSessionDelete, []byte(id))
+		} else {
+			store.baseStore.stores.EventualEventer.AddEventualEvent(EventualEventApiSessionDelete, []byte(id))
 		}
 	}
 
--- a/controller/persistence/eventual_eventer.go
+++ b/controller/persistence/eventual_eventer.go
@@ -95,6 +95,9 @@ type EventualEventRemoved struct {
 type EventualEventProcessingStart struct {
 	// Id is a unique id for processing run
 	Id string
+
+	// StartTime is the time the processing began
+	StartTime time.Time
 }
 
 type EventualEventProcessingBatchStart struct {
@@ -109,6 +112,9 @@ type EventualEventProcessingBatchStart struct {
 
 	// BatchSize is the batch size for the current batch (the maximum value of Count)
 	BatchSize int
+
+	// StartTime the time when the batch started processing
+	StartTime time.Time
 }
 
 type EventualEventProcessingListenerStart struct {
@@ -129,6 +135,12 @@ type EventualEventProcessingListenerStart struct {
 
 	// TotalEventIndex is the total index across all batches of the currently executing event
 	TotalEventIndex int64
+
+	// EventType is the typeof the event that is triggering the listener
+	EventType string
+
+	// StartTime is the time when the listener was started
+	StartTime time.Time
 }
 
 type EventualEventProcessingListenerDone struct {
@@ -152,6 +164,15 @@ type EventualEventProcessingListenerDone struct {
 
 	// Error is nil if no error occurred during execution, otherwise an error value
 	Error error
+
+	// EventType is the typeof the event that triggered the listener
+	EventType string
+
+	// StartTime is the time when the listener started execution
+	StartTime time.Time
+
+	// EndTime is the time when the listener ended execution
+	EndTime time.Time
 }
 
 type EventualEventProcessingBatchDone struct {
@@ -166,6 +187,12 @@ type EventualEventProcessingBatchDone struct {
 
 	// BatchSize is the batch size for the current batch (the maximum value of Count)
 	BatchSize int
+
+	// StartTime the time the batch was started
+	StartTime time.Time
+
+	// EndTime the time the batch ended
+	EndTime time.Time
 }
 
 type EventualEventProcessingDone struct {
@@ -180,6 +207,12 @@ type EventualEventProcessingDone struct {
 
 	// TotalListenersExecuted is the total number of listeners executed during processing
 	TotalListenersExecuted int64
+
+	// StartTime is the time when the processing began
+	StartTime time.Time
+
+	// EndTime is the time when the processing ended
+	EndTime time.Time
 }
 
 const (
@@ -509,8 +542,11 @@ func (a *EventualEventerBbolt) process() {
 		processId: cuid.New(),
 	}
 
+	startTime := time.Now()
+
 	a.Emit(EventualEventProcessingStartName, &EventualEventProcessingStart{
-		Id: info.processId,
+		Id:        info.processId,
+		StartTime: startTime,
 	})
 
 	defer func() {
@@ -521,6 +557,8 @@ func (a *EventualEventerBbolt) process() {
 			TotalBatches:           info.totalBatches,
 			TotalEvents:            info.totalEvents,
 			TotalListenersExecuted: info.totalListenersExecuted,
+			StartTime:              startTime,
+			EndTime:                time.Now(),
 		})
 	}()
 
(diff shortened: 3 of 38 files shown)


controller/model/asynchronous_eventer.go
```diff
@@ -0,0 +1,280 @@
...
+// AsynchronousEventerBbolt implements AsynchronousEventer with a bbolt back storage mechanism.
+// Work is performed on a configurable basis via the Interval property in FIFO order.
+//
+// Events are stored in the following format:
+//		id   - CUID   - a monotonic reference id
+//      name - string - an event name, used for log output
+//      data - []byte - a string array of arguments
+type AsynchronousEventerBbolt struct {
+	handlerMap  cmap.ConcurrentMap //eventName -> handler func(name
+	BucketName  string
```
plorenz, line 78: BucketName is unused, remove?

controller/persistence/asynchronous_event_store.go
```diff
@@ -0,0 +1,99 @@
...
+	entity.LoadBaseValues(bucket)
+	entity.Type = bucket.GetStringOrError(FieldAsynchronousEventType)
+	entity.Data = bucket.Get([]byte(FieldAsynchronousEventData))
+
+}
+
+func (entity *AsynchronousEvent) SetValues(ctx *boltz.PersistContext) {
+	entity.SetBaseValues(ctx)
+	ctx.SetString(FieldAsynchronousEventType, entity.Type)
+	_ = ctx.Bucket.Put([]byte(FieldAsynchronousEventData), entity.Data)
```
plorenz, line 46: `ctx.Bucket.SetError(ctx.Bucket.Put([]byte(FieldAsynchronousEventData), entity.Data))`

controller/persistence/asynchronous_event_store.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/openziti/foundation/storage/ast"
+	"github.com/openziti/foundation/storage/boltz"
+	"go.etcd.io/bbolt"
+)
+
+const (
+	FieldAsynchronousEventType = "type"
+	FieldAsynchronousEventData = "data"
+)
+
```
plorenz, line 29: Wouldn't hate it if we used AsyncEvent instead :) I don't think we lose any clarity with the shorter name

controller/model/asynchronous_eventer.go
```diff
@@ -0,0 +1,280 @@
...
+
+func (a *AsynchronousEventerBbolt) Stop() error {
+	if a.running.CompareAndSwap(true, false) {
+		close(a.stopNotify)
+		return nil
+	}
+
+	return errors.New("not started")
+}
+
```
plorenz, line 171: An alternate approach would be to have a notification channel with space for 1 struct{}, which the run could listen to. Then the Trigger could try to push something onto the channel. If it's already full, then the process is already running. 
The main advantage to this approach is that you can trigger and not worry about the calling thread having to do processing. Thoughts?

controller/model/asynchronous_eventer.go
```diff
@@ -0,0 +1,280 @@
...
+		ctx := boltz.NewMutateContext(tx)
+		return a.env.GetStores().AsynchronousEvent.DeleteById(ctx, id)
+	})
+}
+
+func (a *AsynchronousEventerBbolt) process() {
+	if !a.processing.CompareAndSwap(false, true) {
+		return //already running
+	}
+
```
plorenz, line 192: Do we need this defer if we have the defer where the handlers are called below?

controller/model/asynchronous_eventer.go
```diff
@@ -0,0 +1,280 @@
...
+								pfxlog.Logger().WithError(err).Error("panic caught during asynchronous event %s handler %v", event.Type, handler)
+							}
+						}()
+
+						handler(event.Type, event.Data)
+					}()
+				}
+
+				if err := a.deleteEvent(event.Id); err != nil {
```
plorenz, line 272: Been trying to move any variables in logging into fields, for ease of parsing

controller/persistence/asynchronous_event_store.go
andrewpmartinez: Done

controller/persistence/asynchronous_event_store.go
andrewpmartinez: Was going to change it to "EventualEventer" because it isn't 100% async. But year the name is long.

controller/model/asynchronous_eventer.go
andrewpmartinez: Got it. Fixed

controller/model/asynchronous_eventer.go
andrewpmartinez: We don't need the panic resolution but I do want to make sure processing is set to false.

Removed.

controller/model/asynchronous_eventer.go
andrewpmartinez: What happens if 3 processes all attempt to call `Trigger()` at once between interval runs? The first would start the process the next 2 would block on trying to send to the channel. I don't know if blocking on `Trigger()` is going to be a good idea.

plorenz: You do something like
```
select {
   case trigger <- struct{}:
   default:
}
```
If you can't submit to the trigger channel then someone has already done so and you can continue on your merry way.

andrewpmartinez: I'm ambivalent towards this - I think it will be fine either way. So I will implement the channel.

It does have the side effect of allowing 1 `Trigger` to be requested - which depending on when it is called may immediately pick up new events or cause a query for 0 events.

controller/persistence/session_store.go
```diff
@@ -179,7 +179,7 @@ func (store *sessionStoreImpl) initializeLocal() {
 	store.symbolServicePolicies = store.AddFkSetSymbol(FieldSessionServicePolicies, store.stores.servicePolicy)
 	store.AddSymbol(FieldSessionType, ast.NodeTypeString)
 
-	store.AddFkConstraint(store.symbolApiSession, false, boltz.CascadeDelete)
+	//store.AddFkConstraint(store.symbolApiSession, false, boltz.CascadeNone)
```
andrewpmartinez, line 182: If this is set to `CascadeNone` apiSession deletes will fail w/ a reference error.

Either we just allow any junk data in the field (see session_store_test.go changes) or add a new type of cascade that ignores reference errors on delete.

tests/api_error_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build apitests
```
andrewpmartinez, line 1: Not sure why the golang/goland tooling added this. I sure didn't add it manually.

Need to look into it.

controller/persistence/eventual_eventer.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func (a *EventualEventerBbolt) process() {
+	var count int64 = 0
+	var ids []string
+	var events []*EventualEvent
+	err := a.dbProvider.GetDb().View(func(tx *bbolt.Tx) error {
+
+		var err error
+		ids, count, err = a.store.QueryIds(tx, "limit 50")
```
andrewpmartinez, line 199: Why aren't you using batch size?

andrewpmartinez: SIgh, I'll fix it.

controller/persistence/eventual_eventer.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func (a *EventualEventerBbolt) deleteEvent(id string) error {
+	return a.dbProvider.GetDb().Update(func(tx *bbolt.Tx) error {
+		ctx := boltz.NewMutateContext(tx)
+		return a.store.DeleteById(ctx, id)
+	})
+}
+
+func (a *EventualEventerBbolt) process() {
```
andrewpmartinez, line 192: add metric  gauge for event queue level
add runtime for handlers

controller/persistence/session_store.go
andrewpmartinez: Per conversation add `boltz.CascadeCreateUpdate`

controller/persistence/api_session_store.go
```diff
@@ -109,6 +116,48 @@ type apiSessionStoreImpl struct {
 	symbolIdentity boltz.EntitySymbol
 }
 
+func (store *apiSessionStoreImpl) onDeleteToEventual(i ...interface{}) {
```
andrewpmartinez, line 119: Change to execute as `DeleteById()`  and pass a tx to `EventualEventer.AddEvent(tx, name, []byte(id))`

controller/persistence/api_session_store.go
```diff
@@ -109,6 +116,48 @@ type apiSessionStoreImpl struct {
...
+				"arguments": i,
+			}).
+			WithError(fmt.Errorf("event argument[0] has wrong type %T", i[0])).
+			Error("could not convert argument[0] to *ApiSession")
+	}
+
+}
+
+func (store *apiSessionStoreImpl) onEventualDelete(name string, data []byte) {
+	err := store.stores.DbProvider.GetDb().Update(func(tx *bbolt.Tx) error {
```
andrewpmartinez, line 146: Split to query for ids then delete.

controller/model/testing.go
```diff
@@ -120,20 +120,18 @@ func (ctx *TestContext) GetFingerprintGenerator() cert.FingerprintGenerator {
...
-
-func NewTestContextWithDb(t *testing.T, dbPath string) *TestContext {
 	context := &TestContext{
 		TestContext:     persistence.NewTestContext(t),
 		metricsRegistry: metrics.NewRegistry("test", nil),
 		hostController: &testHostController{
 			closeNotify: make(chan struct{}),
 		},
 	}
-	context.InitWithDbFile(dbPath)
```
andrewpmartinez, line 134: This line was causing init to be called twice in the session enforcer tests (the only user of this function atm) and was causing a double initialization of the underlying stores. The direct result was two EventualEventers running at once which was very confusing to debug. 

Either way, this change makes it so that the TestContext is set up like all the other ones: 1) create via New 2) Init() 3) defer Cleanup().

controller/persistence/api_session_store.go
andrewpmartinez: Fixed

controller/persistence/eventual_eventer.go
andrewpmartinez: done

controller/server/controller.go
```diff
@@ -247,6 +247,10 @@ func (c *Controller) Initialize() {
 
 	}
 
+	if err := c.AppEnv.GetStores().EventualEventer.Start(c.AppEnv.GetHostController().GetCloseNotifyChannel()); err != nil {
+		log.WithError(err).Panic("could not start EventualEventer")
```
dovholuknf, line 251: what are the chances of this? slim to none?

controller/persistence/session_store_test.go
```diff
@@ -52,16 +52,16 @@ func (ctx *TestContext) testCreateInvalidSessions(_ *testing.T) {
 	service := ctx.RequireNewService("test-service")
 
 	session := NewSession("", service.Id)
-	err := ctx.Create(session)
-	ctx.EqualError(err, "fk constraint on sessions.apiSession does not allow null or empty values")
+	//	err := ctx.Create(session)
```
dovholuknf, line 55: dead code in a test? on purpose ? if actually dead let's just remove?

controller/persistence/session_store_test.go
```diff
@@ -34,7 +34,7 @@ func Test_SessionStore(t *testing.T) {
 	ctx.Init()
 
 	t.Run("test create invalid sessions", ctx.testCreateInvalidSessions)
-	t.Run("test update invalid sessions", ctx.testUpdateInvalidSessions)
+	//t.Run("test update invalid sessions", ctx.testUpdateInvalidSessions)
```
dovholuknf, line 37: disabled on purpose?

controller/persistence/eventual_eventer.go
```diff
@@ -0,0 +1,607 @@
...
+
+		a.Emit(EventualEventProcessingDoneName, &EventualEventProcessingDone{
+			Id:                     processId,
+			TotalBatches:           totalBatches,
+			TotalEvents:            totalEvents,
+			TotalListenersExecuted: totalListenersExecuted,
+		})
+	}()
+
+	for {
```
dovholuknf, line 484: lots of code in the while loop here makes it hard to see where it exits.

controller/persistence/eventual_eventer.go
```diff
@@ -0,0 +1,607 @@
...
+		batchEventIndex := int64(0)
+		for _, eventualEvent := range eventualEvents {
+			batchEventIndex++
+			totalEventIndex++
+
+			if handlerVals, found := a.handlerMap.Get(eventualEvent.Type); found {
+				handlers := handlerVals.([]EventListenerFunc)
+				for _, handler := range handlers {
+					func() {
+						listenerExecId := cuid.New()
```
dovholuknf, line 540: why a lambda and not just some function to call?

controller/model/session_handlers.go
```diff
@@ -325,6 +327,47 @@ func (handler *SessionHandler) ListSessionsForEdgeRouter(edgeRouterId string) (*
...
 }
 
+func (handler *SessionHandler) OnApiSessionDelete(name string, data []byte) {
+	apiSessionId := string(data)
+
+	logger := pfxlog.Logger().WithFields(map[string]interface{}{
+		"event":        name,
+		"apiSessionId": apiSessionId,
+	})
+
```
plorenz, line 337: Can we have the scan and delete be in separate transactions, please? Also, you have it in a loop but don't have a limit on the scan.

controller/server/controller.go
andrewpmartinez: Slim to none - only happens on startup

controller/model/session_handlers.go
plorenz: Oh, is this not used anymore, replaced by the method in ApiSessionStore? If that's the case, can probably delete this

controller/persistence/session_store_test.go
andrewpmartinez: Nope, lever overs. Re-enabled here soon.

controller/persistence/eventual_eventer.go
```diff
@@ -0,0 +1,607 @@
...
+		batchSize:         batchSize,
+		trigger:           make(chan struct{}, 1),
+		handlerMap:        cmap.New(),
+		outstandingEvents: &outstanding,
+	}
+}
+
+func (a *EventualEventerBbolt) AddEventualEventWithCtx(ctx boltz.MutateContext, eventType string, data []byte) {
+	newId := cuid.New()
+	total := atomic.AddInt64(a.outstandingEvents, 1)
```
plorenz, line 282: I probably would have put the metric right in the store, but this should be ok too

controller/persistence/eventual_eventer.go
```diff
@@ -0,0 +1,607 @@
...
+
+			if err != nil {
+				pfxlog.Logger().WithError(err).Error("error during EventualEventerBbolt processing, could not query events")
+			}
+
+			if numIds > 0 {
+				break
+			}
+
+			time.Sleep(500 * time.Microsecond)
```
dovholuknf, line 506: eh...  so this for loop could block for up to 2.5s? there's only one running right so shouldn't be a big deal? maybe a comment in here?

plorenz (approved):

controller/persistence/eventual_eventer.go
```diff
@@ -0,0 +1,607 @@
...
+			numEvents = atomic.LoadInt64(a.outstandingEvents)
+		}
+
+		if numIds == 0 {
+			return
+		}
+
+		totalBatches++
```
dovholuknf, line 514: any metrics around total numIds or total time for these for loops to run seem useful?

controller/persistence/session_store_test.go
andrewpmartinez: Fixed, uncommented.

controller/model/session_handlers.go
andrewpmartinez: This isn't used anymore. This is the old version and I never deleted it.

andrewpmartinez: Removed.

controller/persistence/eventual_eventer.go
andrewpmartinez: Events are emitted and metrics can be logged in those event handlers (see appenv.go changes in this PR). I don't know what metrics we care about so I only instrumented the total number of outstanding events.

controller/persistence/eventual_eventer.go
andrewpmartinez: This is dealing with a situation where we know events have been added but the bbolt transaction that added them hasn't been completed yet.

There is a comment, at the top of the loop.

controller/persistence/eventual_eventer.go
andrewpmartinez: defer'ed panic handler - so that an event handler can't kill the eventual eventer.

controller/persistence/eventual_eventer.go
andrewpmartinez: It is 500 microseconds, not milliseconds. At most, it will wait 2,500 microseconds aka 2.5 milliseconds.

dovholuknf (approved):

controller/persistence/eventual_eventer.go
andrewpmartinez: Moved to a separate named function w/ comment.

controller/persistence/eventual_eventer.go
andrewpmartinez: Split up just for you.

PR #861, opened 2022-01-06
adds metrics for endpoints with keys of path[method]

andrewpmartinez:
Example: `/edge/management/v1/authenticate[POST]`


dovholuknf (approved):

plorenz:
We use dots and underscores in metrics names everywhere else. Not sure if this will break anything in the metrics ingest path to elastic. Can we switch it to path.method?

andrewpmartinez:
> We use dots and underscores in metrics names everywhere else. Not sure if this will break anything in the metrics ingest path to elastic. Can we switch it to path.method?

Sure can.

plorenz (approved):

PR #865, opened 2022-01-17
moves metric endpoint logging lower in the stack

andrewpmartinez:
- endpoint url/path -> name remains the same
- start timing on request context before we start processing
- log metrics in IsAllowed which is called by all endpoint after routing
- logging before routing to a handler would log all probe attacks

### DIFF
--- a/controller/env/appenv.go
+++ b/controller/env/appenv.go
@@ -509,8 +509,14 @@ func (ae *AppEnv) IsAllowed(responderFunc func(ae *AppEnv, rc *response.RequestC
 
 		responderFunc(ae, rc)
 
-		timer := ae.GetHostController().GetNetwork().GetMetricsRegistry().Timer(getMetricTimerName(rc.Request))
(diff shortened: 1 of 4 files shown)


controller/env/appenv.go
```diff
@@ -485,6 +508,15 @@ func (ae *AppEnv) IsAllowed(responderFunc func(ae *AppEnv, rc *response.RequestC
 		}
 
 		responderFunc(ae, rc)
+
+		if !rc.StartTime.IsZero() {
+			timer := ae.GetHostController().GetNetwork().GetMetricsRegistry().Timer(getMetricTimerName(rc.Request))
```
andrewpmartinez, line 513: Moved here because this function (`IsAllowed`) is only invoked after the REST API has routed the HTTP request to a known handler. Unknown handlers return a 404 before they reach this location.

dovholuknf (approved):

plorenz (approved):

PR #868, opened 2022-01-19
adds identity specific REST API metrics

andrewpmartinez:


controller/env/appenv.go
```diff
@@ -541,8 +541,18 @@ func (ae *AppEnv) IsAllowed(responderFunc func(ae *AppEnv, rc *response.RequestC
...
-			timer.UpdateSince(rc.StartTime)
+			urlMetricName := getMetricTimerName(rc.Request)
+
+			//endpoint metrics
+			urlTimer := ae.GetMetricsRegistry().Timer(urlMetricName)
+			urlTimer.UpdateSince(rc.StartTime)
+
+			//endpoint per identity metrics
+			if rc.Identity != nil {
+				urlIdentityTimer := ae.GetMetricsRegistry().Timer("identity." + rc.Identity.Id + "." + urlMetricName)
```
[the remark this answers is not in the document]
andrewpmartinez: Do you simply want it as <url>.<method>.identity.<identity-id> ?

controller/env/appenv.go
```diff
@@ -541,8 +541,18 @@ func (ae *AppEnv) IsAllowed(responderFunc func(ae *AppEnv, rc *response.RequestC
...
 
 		if !rc.StartTime.IsZero() {
-			timer := ae.GetHostController().GetNetwork().GetMetricsRegistry().Timer(getMetricTimerName(rc.Request))
-			timer.UpdateSince(rc.StartTime)
+			urlMetricName := getMetricTimerName(rc.Request)
+
+			//endpoint metrics
+			urlTimer := ae.GetMetricsRegistry().Timer(urlMetricName)
+			urlTimer.UpdateSince(rc.StartTime)
+
```
plorenz, line 549: Do we want to do these as events, rather than as a metric? I'm concerned that as the number of identities scales up, this will become a memory issue. If we do them as events, they are transient. Also, any given identity should only have a limited number of events. We could also introduce some back-pressure and drop events if we can't keep up.

andrewpmartinez: I'm fine either way, but what you are suggesting means this isn't going in today.

PR #892, opened 2022-02-09
alters posture query updateAt to also indicate query state changes

andrewpmartinez:
- formerly posture queries only had an updatedAt time copied from the
  source posture check which. For MFA posture checks this is only
  correct when not using promptOnWake/Unlock
- prompt timer reductions due to wake/unlock now update postureQuery
  updatedAt values
- subsequent reconfigurations to the posture check (i.e. patching a
  posture check) is also taken into account - latest time.Time wins
- test for MFA posture check no longer use gabs and instead use the
  relevant input/output rest_model.* structures

### DIFF
--- a/controller/internal/routes/service_api_model.go
+++ b/controller/internal/routes/service_api_model.go
@@ -194,6 +194,13 @@ func MapServiceToRestModel(ae *env.AppEnv, rc *response.RequestContext, service 
 			timeout := postureCheck.TimeoutSeconds()
 			timeoutRemaining := postureCheck.TimeoutRemainingSeconds(rc.ApiSession.Id, ae.Handlers.PostureResponse.PostureData(rc.Identity.Id))
 
+			//determine if the last updated at is determined the posture checks or posture query to signal state change
+			if lastUpdatedAt := postureCheck.LastUpdateAt(rc.ApiSession.Id, ae.Handlers.PostureResponse.PostureData(rc.Identity.Id)); lastUpdatedAt != nil {
+				if lastUpdatedAt.After(postureCheck.UpdatedAt) {
+					query.UpdatedAt = DateTimePtrOrNil(lastUpdatedAt)
+				}
+
+			}
 			query.IsPassing = &isCheckPassing
 			query.TimeoutRemaining = &timeoutRemaining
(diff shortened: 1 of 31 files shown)


controller/model/posture_check_model_mac.go
```diff
@@ -20,13 +20,19 @@ import (
...
 )
 
 var _ PostureCheckSubType = &PostureCheckMacAddresses{}
 
 type PostureCheckMacAddresses struct {
 	MacAddresses []string
 }
+
+func (p *PostureCheckMacAddresses) LastUpdatedAt(apiSessionId string, pd *PostureData) *time.Time {
+	return nil
```
dovholuknf, line 33: nil on purpose? feels like this needs 'a comment'?

controller/model/posture_check_model_os.go
```diff
@@ -32,6 +33,10 @@ type PostureCheckOperatingSystem struct {
 	OperatingSystems []OperatingSystem
 }
 
+func (p *PostureCheckOperatingSystem) LastUpdatedAt(id string, pd *PostureData) *time.Time {
+	return nil
```
dovholuknf, line 37: same comment - won't make it again should it reoccur. :)

dovholuknf (approved):

controller/model/posture_check_model_mac.go
andrewpmartinez: It is an interface function, the documentation is on the interface.

controller/internal/routes/service_api_model.go
```diff
@@ -194,6 +194,13 @@ func MapServiceToRestModel(ae *env.AppEnv, rc *response.RequestContext, service
 			timeout := postureCheck.TimeoutSeconds()
 			timeoutRemaining := postureCheck.TimeoutRemainingSeconds(rc.ApiSession.Id, ae.Handlers.PostureResponse.PostureData(rc.Identity.Id))
 
+			//determine if the last updated at is determined the posture checks or posture query to signal state change
```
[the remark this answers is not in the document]
andrewpmartinez: Yea I missed a word or two.

PR #899, opened 2022-02-21
Router cost

camotts:
Handled router cost changes from fabric in edge


specs/management.yml
```diff
@@ -12226,6 +12243,11 @@ definitions:
     properties:
       appData:
         $ref: '#/definitions/tags'
+      cost:
```
plorenz, line 12246: cost should be required in edge router update. Unlike in the fabric, fingerprint should not be updateable here, since there's a separate process for managing enrollments/fingerprints that we don't want to bypass.

specs/management.yml
```diff
@@ -13884,6 +13921,11 @@ definitions:
     required:
     - name
     properties:
+      cost:
```
plorenz, line 13924: cost required here as well

plorenz (changes requested):
Overall looks good. Couple of places in the spec where cost should be required for updates. Also, please revert any files that are just import organization.

specs/management.yml
camotts: added

specs/management.yml
camotts: added

plorenz (approved):
Looks good!

It's not required, but re-requesting review from the person who requested changes when changes are made lets them know to take another look.

PR #903, opened 2022-02-23
adds cert authenticator extension endpoint

andrewpmartinez:
- adds endpoint POST /edge/management/v1/authenticator/{id}/extend
- adds endpoint POST /edge/management/v1/authenticator/{id}/extend-verify
- adds endpoint POST /edge/client/v1/authenticator/{id}/extend
- adds endpoint POST /edge/client/v1/authenticator/{id}/extend-verify
- add logic that allows a client endpoint to extend the valid period for
  their current certificate authenticator
- requires clients to use their existing client cetificate for extension
- requires clients to verify they have received the new public key
- allows clients to use a new private key if desired
- allows only 1st party certificate authenticators to extend


plorenz (approved):
Looks good. Noticed a test stub at the end, not sure of that was intentional.

PR #904, opened 2022-02-24
Separate fabric and edge entity counts for parent/child stores

plorenz:


controller/persistence/stores.go
```diff
@@ -76,7 +76,41 @@ func (stores *Stores) buildStoreMap() {
...
 }
 
-func (stores *Stores) GetStoreList() []Store {
+func (stores *Stores) GetEntityCounts(dbProvider DbProvider) (map[string]int64, error) {
+	result := map[string]int64{}
+	for _, store := range stores.storeMap {
+		err := dbProvider.GetDb().View(func(tx *bbolt.Tx) error {
+			key := store.GetEntityType()
+			if store.IsChildStore() {
+				if store.GetEntityType() == "routers" {
```
andrewpmartinez, line 85: do we have constants for the types?

andrewpmartinez (approved):

controller/persistence/stores.go
plorenz: Fixed

PR #906, opened 2022-02-28
Added global constants files for use in this and other projects

gberl002:


controller/constants.go
```diff
@@ -0,0 +1,23 @@
...
+	DefaultEdgeSessionTimeout = 10 * time.Minute
+	MinEdgeSessionTimeout     = 1 * time.Minute
+
+	MinEdgeEnrollmentDuration     = 5 * time.Minute
+	DefaultEdgeEnrollmentDuration = 5 * time.Minute
+
+	DefaultHttpIdleTimeout       = 5000 * time.Millisecond
+	DefaultHttpReadTimeout       = 5000 * time.Millisecond
+	DefaultHttpReadHeaderTimeout = 5000 * time.Millisecond
+	DefaultHttpWriteTimeout      = 100000 * time.Millisecond
```
andrewpmartinez, line 22: Why put them so far from where they are used?

andrewpmartinez (changes requested):
See comment.

controller/constants.go
andrewpmartinez: Moving the file to `config` is the same (to external code) as not changing it at all.

gberl002: @andrewpmartinez so are you okay with moving the file to the config package or are you saying you want the constants to go in config.go since it's essentially the same due to Go organization with packages?

andrewpmartinez: The later. golang doesn't care about individual files just folder names.

Having the constants near the code in individual files in `/config/*.go`  vs `/config/config.go` is the same for external code. It will be `config.<constant>`. In that vein, this PR can be abandoned.

dovholuknf: i disagree that this adds any 'distance' from the code and the constant. i'd much rather have the constants in a separate file.  but whatever ¯\_(ツ)_/¯

andrewpmartinez: This is a change for change's sake. Not for any empirical reason. The external code doesn't care which files the constants are in.

For the internal code, we can simply delete 1 file and be done and not worry about leftover unused constants.

I do agree it is "shruggable" but when if there is no clear gain, I don't see the point.

controller/config/config.go
```diff
@@ -65,9 +77,9 @@ type Api struct {
 }
 
 type Config struct {
-	Enabled            bool
-	Api                Api
-	Enrollment         Enrollment
```
gberl002, line 70: I'm not sure why these are showing as changed, even though it's just a formatting change, they don't show as being changed on my local files.

andrewpmartinez: It is a white space change, just shifted them all to the left. Should be fine.

andrewpmartinez (approved):

PR #910, opened 2022-03-03
Updating test to work with the latest global config constants

gberl002:
var type was changed from int to time.duration

### DIFF
--- a/router/xgress_edge/factory_test.go
+++ b/router/xgress_edge/factory_test.go
@@ -1,11 +1,12 @@
 package xgress_edge
 
 import (
+	"github.com/openziti/channel"
 	"github.com/openziti/edge/edge_common"
 	"github.com/openziti/fabric/router/xgress"
-	"github.com/openziti/channel"
 	"github.com/stretchr/testify/require"
 	"testing"
+	"time"
 )
 
 /*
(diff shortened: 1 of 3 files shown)


router/xgress_edge/factory_test.go
```diff
@@ -107,7 +108,7 @@ func Test_load(t *testing.T) {
 const (
 	connectMaxQueuedConnections   = 50
 	connectMaxOutstandingConnects = 100
-	connectTimeoutMs              = 3000
```
andrewpmartinez, line 110: Did you intend to change the value from 3s to 1s?

router/xgress_edge/factory_test.go
```diff
@@ -107,7 +108,7 @@ func Test_load(t *testing.T) {
...
+	connectTimeoutMs              = 1000 * time.Millisecond
 )
```
gberl002, line 112: I honestly cannot figure out why this was previously 3000 and passed. The code in the channel project had an int value of 1000. That value of 1000 was changed to a time.duration value do I've updated that here.

andrewpmartinez (approved):

router/xgress_edge/factory_test.go
gberl002: Yes, this is what I commented on below. I don't know how it was passing before as the value it is compared against was 1000. I looked through the commit history and it has always been 1000 since the creation of the file 3 years ago... very strange.

PR #933, opened 2022-03-18
Route error codes

camotts:


router/xgress_edge_tunnel/dialer.go
```diff
@@ -51,7 +52,7 @@ func (self *tunneler) Dial(destination string, circuitId *identity.TokenId, addr
```
plorenz, line 52: Above, where `self.terminators.get(destination)` returns an error if the terminator is not found is where we'd want to return an InvalidTerminator error

router/xgress_edge/dialer.go
```diff
@@ -124,7 +125,7 @@ func (dialer *dialer) Dial(destination string, circuitId *identity.TokenId, addr
 		if err != nil {
```
plorenz, line 125: We'd also want to return an InvalidTerminator error above, if `listenConn, found := dialer.factory.hostedServices.Get(token)` doesn't find the terminator.

router/xgress_edge/dialer.go
```diff
@@ -124,7 +125,7 @@ func (dialer *dialer) Dial(destination string, circuitId *identity.TokenId, addr
...
 			conn.close(false, err.Error())
 			x.Close()
```
plorenz, line 127: This one should just be a regular dial error. Terminator exists, but had trouble dialing.

router/xgress_edge_transport/dialer.go
```diff
@@ -52,13 +54,13 @@ func (txd *dialer) Dial(destination string, circuitId *identity.TokenId, address
 
 	txDestination, err := transport.ParseAddress(destination)
 	if err != nil {
-		return nil, fmt.Errorf("cannot dial on invalid address [%s] (%s)", destination, err)
+		return nil, xgress.InvalidTerminatorError{InnerError: fmt.Errorf("cannot dial on invalid address [%s] (%s)", destination, err)}
 	}
 
 	log.Debug("dialing")
 	peer, err := txDestination.Dial("x/"+circuitId.Token, circuitId, txd.options.ConnectTimeout, nil)
 	if err != nil {
```
plorenz, line 62: Should be a regular dial error

router/xgress_edge_tunnel/dialer.go
```diff
@@ -51,7 +52,7 @@ func (self *tunneler) Dial(destination string, circuitId *identity.TokenId, addr
 
 	conn, halfClose, err := terminator.context.Dial(options)
 	if err != nil {
```
plorenz, line 54: Should be a regular dial error

router/xgress_edge_transport/dialer.go
camotts: This is how I am checking back in the controller if its an "expected" error. The dial error gets persisted in the inner error and checked on the controller side. Does that need to be redone?

plorenz (approved):

PR #935, opened 2022-03-21
adds auth policy CRUD, does not affect authentication

andrewpmartinez:


go.work
```diff
@@ -0,0 +1,3 @@
+go 1.17
```
plorenz, line 1: go.work shouldn't be committed. We should add it to .gitignore

andrewpmartinez: Woops, I was playing around with it and I forgot it was there.

plorenz (approved):

PR #946, opened 2022-03-28
adds auth policy / disabled identity auth logic

andrewpmartinez:


controller/model/authenticator_mod_ext_jwt.go
```diff
@@ -67,31 +70,39 @@ func (a *AuthModuleExtJwt) pubKeyLookup(token *jwt.Token) (interface{}, error) {
 	kidVal, ok := token.Header["kid"]
 
 	if !ok {
+		pfxlog.Logger().Error("missing kid")
```
qrkourier, line 73: saddest log message ever :cry:

andrewpmartinez: Haha, reading that out of context only made me second guess why that message exists.

controller/internal/routes/identity_router.go
```diff
@@ -150,6 +150,18 @@ func (r *IdentityRouter) Register(ae *env.AppEnv) {
 		}, params.HTTPRequest, params.ID, "", permissions.IsAdmin())
 	})
```
plorenz, line 152: Why have explicit operations for these, instead of just a patch on the enabled/disabled flag? 

I don't think this is wrong, just curious as to why this is your preference

andrewpmartinez: The bool flag is virtual. It is derived from two other fields: `disabledAt` and `disabledUntil`. Identities can be locked indefinitely by not having `disabledUntil`  set.

Disabling is done by making a disable request with a duration or no duration (indefinite). Enabling is clearing both fields.

It is easier to define this behavior with a specific endpoint rather than updating two fields in concert.

plorenz: Makes sense, I just noticed the disabledUntil as I was further down.

plorenz (approved):

controller/model/identity_handlers.go
```diff
@@ -501,6 +502,76 @@ func (handler *IdentityHandler) VisitIdentityAuthenticatorFingerprints(tx *bbolt
 	return stopVisit, err
 }
```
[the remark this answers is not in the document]
andrewpmartinez: Allows users to add external ids (i.e. usually emails, employee numbers, etc) that are used to identify the owner of an identity from a separate system.

Necessary when consuming JWTs from external systems where their ID doesn't or can't match Ziti's internal id. Consider a JWT that puts an email in the `sub` field. We need to a way to match that `sub` to an identity for authentication. Enter `externalId`.

I didn't want `externalId` to be on tags because I wanted it to be a separate field that is named and has an index for lookups.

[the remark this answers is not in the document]
plorenz: Oh, I was just wondering if the spelling was wrong :) You've got Ready instead of Read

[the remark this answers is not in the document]
andrewpmartinez: OH. Yea I'll fix that.

PR #953, opened 2022-03-31
add error return on err in cert auth, update db version

andrewpmartinez:


controller/persistence/migrations.go
```diff
@@ -22,7 +22,7 @@ import (
 )
 
 const (
```
plorenz, line 24: Need to fix the version check below too

andrewpmartinez: Oh I see

plorenz (approved):

controller/persistence/migrations.go
andrewpmartinez: fixed

PR #963, opened 2022-04-08
fixes is admin not set on updates (put/patch)

andrewpmartinez:


controller/internal/routes/identity_api_model.go
```diff
@@ -33,8 +33,7 @@ import (
 )
 
 const (
-	EntityNameIdentity              = "identities"
-	EntityNameIdentityServiceConfig = "service-configs"
```
andrewpmartinez, line 37: This was unused

plorenz (approved):

PR #977, opened 2022-04-16
use gotestfmt for nicer test output

plorenz:


andrewpmartinez:
Interesting that it is showing failed tests that I haven't seen failing. It might be all the log spam from the controllers and routers?

plorenz:
gotestfmt not working well enough yet

PR #997, opened 2022-05-05
added small logging on successful terminator removal

camotts:


router/xgress_edge/listener.go
```diff
@@ -97,6 +98,8 @@ func (self *edgeClientConn) HandleClose(_ channel.Channel) {
 	for _, terminator := range terminators {
 		if err := self.removeTerminator(terminator); err != nil {
 			log.Warnf("failed to remove terminator %v for session with token %v on channel close", terminator.terminatorId, terminator.token)
+		} else {
+			log.Infof("successfully removed terminator %v for session with token %v on channel close", terminator.terminatorId, terminator.token)
 		}
```
plorenz, line 103: I've been trying to switch to more structured logging, since it's easier for the tooling to work with. So doing something like:

```
log.WithField("terminatorId", terminatorId).WithField("token", token).Info("log text here)
```
Don't have to change this necessarily, just something to keep in mind for the future.

plorenz (approved):

plorenz (approved):

PR #1016, opened 2022-05-23
adds x509 claims support for external ids to CAs

andrewpmartinez:
- adds REST API CA external id field
- adds rest api for external fields
- refactors authentication interfaces for federation
- alters authentication moduels to return AuthResult interfaces
- allows AuthResults to provide external x509 certs as session certs


controller/model/authenticator_mod_cert_test.go
```diff
@@ -0,0 +1,23 @@
...
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/
+
+package model
+
+import "testing"
+
```
plorenz, line 20: Is this TBD or a forgotten remnant?

plorenz (approved):
Looks good other than the test stub

controller/model/authenticator_mod_cert_test.go
andrewpmartinez: Remnant. Added tests to other files.

Will remove.

PR #1018, opened 2022-05-24
adds utility facilities for creating rest clients for the management api

andrewpmartinez:


rest_util/capool.go
```diff
@@ -0,0 +1,88 @@
...
+	"crypto/x509"
+	"encoding/base64"
+	"fmt"
+	"github.com/fullsailor/pkcs7"
+	"io/ioutil"
+)
+
+// VerifyController will attempt to use the provided x509.CertPool to connect to the provided controller.
+// If successful true an no error will be returned.
+func VerifyController(controllerAddr string, caPool *x509.CertPool) (bool, error) {
```
camotts, line 29: Just an unecessary nit picky question. Will there one day be a scenario where we can expect a `false, nil` return? Otherwise seems a bit redundant to be checking the bool here if they always mean the same thing

camotts (approved):

gberl002 (approved):

rest_util/capool.go
andrewpmartinez: `false` is the zero value of a bool - so there is no way to both return true/false and or nil/not nill errors w/o it being false.

The scenarios are:

- `true, nil` - all is okay
- `true, error` - not possible
- `false, error` - there was a connection error or other runtime error, could not validate the controller and here is the error that cause the issue
- `false, nil` - the controller was contacted and failed to validate, there was no unexpected errors

Issue #1020, opened 2022-05-26
500 in Edge Controller when creating config


gberl002:
I posted this payload to my edge controller. The same thing seems to work from postman but doesn't work from go test using the API.

```
POST /edge/management/v1/configs HTTP/1.1
Host: ziti-edge-controller:1280
User-Agent: Go-http-client/1.1
Content-Length: 166
Accept: application/json
Content-Type: application/json
Zt-Session: xxxsecretxxx
Accept-Encoding: gzip

{"configTypeId":"g7cIWbcGg","data":"'{\"protocols\":[\"tcp\"],\"addresses\":[\"simple.web.test\"], \"portRanges\":[{\"low\":80, \"high\":80}]}'","name":"basic.dial"}
```
This produced a 500 in the controller with the following stack:

```
ERROR fabric/controller/api.(*timeoutHandler).ServeHTTP.func1.1: panic caught by timeout next: interface conversion: interface {} is string, not map[string]interface {}
goroutine 2754202 [running]:
github.com/openziti/foundation/util/debugz.generateStack(0x2000, 0xc0?)
	/root/go/pkg/mod/github.com/openziti/foundation@v0.17.25/util/debugz/stack.go:39 +0x4c
github.com/openziti/foundation/util/debugz.GenerateLocalStack(...)
	/root/go/pkg/mod/github.com/openziti/foundation@v0.17.25/util/debugz/stack.go:34
github.com/openziti/fabric/controller/api.(*timeoutHandler).ServeHTTP.func1.1()
	/root/go/pkg/mod/github.com/openziti/fabric@v0.17.135/controller/api/timeouts.go:75 +0xa8
panic({0x13b49a0, 0x4001339530})
	/usr/local/go/src/runtime/panic.go:838 +0x20c
github.com/openziti/edge/controller/internal/routes.MapCreateConfigToModel(0x40013393b0)
	/root/go/pkg/mod/github.com/openziti/edge@v0.21.213/controller/internal/routes/config_api_model.go:44 +0x184
github.com/openziti/edge/controller/internal/routes.(*ConfigRouter).Create.func1()
	/root/go/pkg/mod/github.com/openziti/edge@v0.21.213/controller/internal/routes/config_router.go:85 +0x2c
github.com/openziti/edge/controller/internal/routes.CreateWithResponder(0x40021962a0, {0x1a56ee8, 0x40021962a0}, {0x1a2c8f8, 0x40004440f0}, 0x40013392f0?)
[...]

PR #1021, opened 2022-05-26
fix scenarios where `nil` or non map is provided in data causes a panic

andrewpmartinez:
- fixes edge #1020

### DIFF
--- a/controller/persistence/config_store.go
+++ b/controller/persistence/config_store.go
@@ -64,7 +64,7 @@ func (entity *Config) SetValues(ctx *boltz.PersistContext) {
 	ctx.SetString(FieldConfigType, entity.Type)
 	ctx.SetMap(FieldConfigData, entity.Data)
 
-	if entity.Data == nil {
+	if ctx.ProceedWithSet(FieldConfigData) && entity.Data == nil {
 		ctx.Bucket.SetError(errorz.NewFieldError("data is required", "data", nil))
 	}
 }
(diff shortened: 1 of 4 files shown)


controller/persistence/config_store.go
```diff
@@ -62,6 +63,10 @@ func (entity *Config) SetValues(ctx *boltz.PersistContext) {
 	ctx.SetString(FieldName, entity.Name)
 	ctx.SetString(FieldConfigType, entity.Type)
 	ctx.SetMap(FieldConfigData, entity.Data)
+
+	if entity.Data == nil {
+		ctx.Bucket.SetError(errorz.NewFieldError("data is required", "data", nil))
```
dovholuknf, line 68: is this accurate? wasn't @gberl002 sending data just that data didn't deserialize properly due to the format? would you consider adding "or data improperly formatted" to this response? (assuming I understand what happened)

dovholuknf (approved):

gberl002 (approved):

controller/persistence/config_store.go
andrewpmartinez: Not at this level.

dovholuknf: not at this level 'in the code' - but somewhere else? I just want the consumer to get a good message to help them understand the problem. As long as a consumer gets a "hey you screwed up" (speaking of which is this now a 400) and 'how' that's all I care deeply about.

andrewpmartinez: https://github.com/openziti/edge/pull/1021/files#diff-aabaaa26930c9df6015bb4c88a9ceae28fafc4e3431545f3c916202c630648cdR50

PR #1030, opened 2022-06-02
Raft

plorenz:


controller/internal/routes/config_router.go
```diff
@@ -119,6 +122,6 @@ func (r *ConfigRouter) Patch(ae *env.AppEnv, rc *response.RequestContext, params
 		if err != nil {
 			return err
 		}
-		return ae.Handlers.Config.Patch(model, fields.FilterMaps("tags", "data"))
+		return ae.Managers.Config.Update(model, fields.FilterMaps("tags", "data"))
```
dovholuknf, line 125: just checking that this was an intentional change. I don't know the differences tween patch/update but this seems like it was a bug that you found/fixed along the way?

plorenz: Only difference between update and patch is that patch allowed a FieldChecker (to specify which fields had changed). Now Update takes a field checker and you can pass in nil if you want to update all fields. This style works a bit better with the update command we have which includes an optional updated fields field.

tests/data_flow_test.go
```diff
@@ -43,7 +47,7 @@ func Test_Dataflow(t *testing.T) {
 
 	testServer := newTestServer(listener, func(conn *testServerConn) error {
 		for {
-			name, eof := conn.ReadString(1024, 1*time.Minute)
+			name, eof := conn.ReadString(math.MaxUint16*4, 1*time.Minute)
```
dovholuknf, line 50: out of curiosity - why MaxUint16 * 4 instead of just "1024"? same comment wrt 512 below vs MaxUint16*2... seems... strange. makes me think there's a reason?

controller/model/identity_handlers.go
```diff
@@ -241,14 +248,34 @@ func (handler *IdentityHandler) InitializeDefaultAdmin(username, password, name
...
-		pfxlog.Logger().Panic(err)
+		return err
 	}
 
 	if identity != nil {
 		return errors.New("already initialized: Ziti Edge default admin already defined")
 	}
 
-	identityType, err := handler.env.GetHandlers().IdentityType.ReadByName(IdentityTypeUser)
+	if len(username) < minDefaultAdminUsernameLength {
```
dovholuknf, line 258: are all these settings configurable? i'm just happy my "dev" password is at least 4 chars long. i almost feel like security software could take the "you really should know better" attitude ? :)

controller/model/create_terminator_cmd.go
```diff
@@ -0,0 +1,103 @@
...
+		return err
+	}
+
+	if terminator.GetIdentity() == "" {
+		return nil
+	}
+
+	identityTerminators, err := self.Env.GetStores().Terminator.GetTerminatorsInIdentityGroup(tx, terminator.GetId())
+	for _, otherTerminator := range identityTerminators {
+		otherSession, err := self.getTerminatorSession(tx, otherTerminator, "sibling ")
```
dovholuknf, line 41: just confirming the space here in "sibling " intentional?

controller/model/create_terminator_cmd.go
```diff
@@ -0,0 +1,103 @@
...
+type terminator interface {
+	GetId() string
+	GetIdentity() string
+	GetBinding() string
+	GetAddress() string
+}
+
+func (self *CreateEdgeTerminatorCmd) getTerminatorSession(tx *bbolt.Tx, terminator terminator, context string) (*persistence.Session, error) {
+	if terminator.GetBinding() != edge_common.EdgeBinding {
+		return nil, errors.Errorf("%vterminator %v with identity %v is not edge terminator. Can't share identity", context, terminator.GetId(), terminator.GetIdentity())
```
dovholuknf, line 64: `%vterminator` should there be a space there? seems like a small issue with this message? ... maybe this is why "sibling " has a space in it (comment from before). i expect that's the case.

controller/model/create_terminator_cmd.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	addressParts := strings.Split(terminator.GetAddress(), ":")
+	if len(addressParts) != 2 {
+		return nil, errors.Errorf("%vterminator %v with identity %v is not edge terminator. Can't share identity", context, terminator.GetId(), terminator.GetIdentity())
+	}
+
+	if addressParts[0] != "hosted" {
+		return nil, errors.Errorf("%vterminator %v with identity %v is not edge terminator. Can't share identity", context, terminator.GetId(), terminator.GetIdentity())
```
dovholuknf, line 73: just an observation but all this code seem like it'd be good in its own function maybe. maybe it's not worth it...

dovholuknf (approved):

controller/model/create_terminator_cmd.go
plorenz: Yeah, used for error messages and it can be either blank or a string. The space just means the target method doesn't need to figure out if a space is necessary

controller/model/create_terminator_cmd.go
plorenz: Yes, correct.

controller/model/managers.go
```diff
@@ -0,0 +1,135 @@
+/*
+	Copyright NetFoundry, Inc.
```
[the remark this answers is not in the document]
plorenz: Fixed, but there's many other others. Should probably run a sed script to fix these

controller/model/identity_handlers.go
plorenz: It's not currently configurable. I was just moving some code around here. I don't know if Andrew has plans to make it configurable at some point.

PR #1032, opened 2022-06-03
Extending default enrollment period and session timeout

gberl002:


controller/config/config.go
```diff
@@ -41,11 +41,11 @@ const (
 	MaxEdgeAPIActivityUpdateInterval      = 10 * time.Minute
 	MinEdgeAPIActivityUpdateInterval      = time.Millisecond
 
-	DefaultEdgeSessionTimeout = 10 * time.Minute
+	DefaultEdgeSessionTimeout = 30 * time.Minute
```
dovholuknf, line 44: add the same comment here please... anything that's used in the config template should get this same comment : `// Don't change the time unit, keep time.Minute`

gberl002:
Having trouble signing previous commits, will close and start a new branch/PR

PR #1034, opened 2022-06-03
Extending default enrollment period and session timeout

gberl002:


controller/config/config.go
```diff
@@ -41,10 +41,10 @@ const (
 	MaxEdgeAPIActivityUpdateInterval      = 10 * time.Minute
 	MinEdgeAPIActivityUpdateInterval      = time.Millisecond
 
-	DefaultEdgeSessionTimeout = 10 * time.Minute
+	DefaultEdgeSessionTimeout = 30 * time.Minute // Don't change the time unit, keep time.Minute
 	MinEdgeSessionTimeout     = 1 * time.Minute
 
-	MinEdgeEnrollmentDuration     = 5 * time.Minute
+	MinEdgeEnrollmentDuration     = 180 * time.Minute // Don't change the time unit, keep time.Minute
```
andrewpmartinez, line 47: The comments make no sense. It isn't a big deal but I would remove them. The unit doesn't matter as long as the duration value that is generated is the desired value.

andrewpmartinez:
There is config in `ziti/etc` that cite the 5minute default, those will also need to be updated.

andrewpmartinez:
Also please don't remove maintainers from your PR

controller/config/config.go
dovholuknf: oh - actually that's a dandy point. as long as the temaplate processes it properly the comments aren't needed. good point...

dovholuknf:
> There is config in `ziti/etc` that cite the 5minute default, those will also need to be updated.

I don't see a need to update that in this PR. I'd prefer those go away entirely myself but we can battle that live if you wish? :)

andrewpmartinez:
> > There is config in `ziti/etc` that cite the 5minute default, those will also need to be updated.
> 
> I don't see a need to update that in this PR. I'd prefer those go away entirely myself but we can battle that live if you wish? :)

It would have to be a separate PR because it is in another repo.

controller/config/config.go
dovholuknf: on re-instpection, this is the `MinEdgeEnrollmentDuration` we want to change `DefaultEdgeEnrollmentDuration`... the minimum can stay at 5m

dovholuknf (approved):

controller/config/config.go
gberl002: The template does use it properly, they were converted to time.durations during the last refactoring of the default values.

PR #1040, opened 2022-06-06
fix regression in jwt auth

rentallect:


controller/model/authenticator_mod_ext_jwt.go
```diff
@@ -112,7 +113,7 @@ type AuthResultJwt struct {
 }
 
 func (a *AuthResultJwt) IsSuccessful() bool {
-	return a.externalJwtSignerId != "" && a.AuthResultBase.identityId != ""
+	return a.externalJwtSignerId != "" && a.AuthResultBase.externalId != ""
```
andrewpmartinez, line 116: This seems fishy.

andrewpmartinez (changes requested):
This needs tests.

controller/model/authenticator_mod_ext_jwt.go
andrewpmartinez: I do think you are onto something though.

rentallect:
It does need tests.  I acknowledge.  The tests that are out there now
fail...keep reading...

The edge tweak I did is now working for browZer.

The edge tests have no notion of an externalId, and I think that is why
this regression was not caught by the tests.

rentallect:
Closing this.  Fix should appear over in https://github.com/openziti/edge/pull/1029

PR #1051, opened 2022-06-08
adds identity enrollments endpoint

andrewpmartinez:
- removes unused details field
- adds tests, adds validation formats


rest_management_api_server/operations/identity/get_identity_enrollments_urlbuilder.go
```diff
@@ -0,0 +1,123 @@
+// Code generated by go-swagger; DO NOT EDIT.
+
+//
+// Copyright NetFoundry, Inc.
```
dovholuknf, line 4: All these should be NetFoundry Inc. not NetFoundry, Inc. please :)

andrewpmartinez: Separate PR because it hits 1.7k files

PR #1057, opened 2022-06-17
allow externalid to be nullable

dovholuknf:


andrewpmartinez (changes requested):
This is a generated file. 

This change should be made in the base file `edge/specs/source/management/identities.yml`

Then the spec needs to be generated which will update all of the generated files.

To generate:


1.install `swagger` v0.28.0 from https://github.com/go-swagger/go-swagger/releases
2. cd to the root `edge` directory,
3. execute: `./scripts/generate_rest.ps1` or `./scripts/generate_rest.sh`

PR #1060, opened 2022-06-21
Fix various races detected by the race detector

plorenz:


tunnel/dns/server.go
```diff
@@ -150,13 +150,22 @@ func (r *resolver) testSystemResolver() error {
...
 }
 
-func (r *resolver) getAddress(name string) (net.IP, error) {
+func (r *resolver) getHostnameIp(name string) (net.IP, bool) {
+	r.namesMtx.Lock()
+	defer r.namesMtx.Unlock()
 	canonical := strings.ToLower(name)
-	a, ok := r.names[canonical]
+	ip, found := r.names[canonical]
+	return ip, found
```
andrewpmartinez, line 158: This will work, but I'm curious to know why a mutex over a self-contained implementation from cmap?

andrewpmartinez (approved):

tunnel/dns/server.go
plorenz: cmap would probably be fine as well, but we already had the mutex there, just not consistently applied, so I went for the minimally intrusive fix

PR #1141, opened 2022-08-11
Cmd encode decode fixes

plorenz:


controller/model/posture_check_manager.go
```diff
@@ -144,7 +145,10 @@ func (self *PostureCheckManager) Marshall(entity *PostureCheck) ([]byte, error)
 		RoleAttributes: entity.RoleAttributes,
 	}
 
-	entity.SubType.fillProtobuf(msg)
+	if entity.SubType != nil {
```
andrewpmartinez, line 148: Interesting, subtype should never be nil.

andrewpmartinez: Unless this is to account for a partial update? Say a name change only?

pb/edge_cmd_pb/impl.go
```diff
@@ -83,78 +87,159 @@ func DecodeTags(tags map[string]*TagValue) map[string]interface{} {
...
+		encoded, err := EncodeJsonValue(v)
+		if err != nil {
+			return nil, err
+		}
+		result[k] = encoded
+	}
+	return &JsonMap{Value: result}, nil
+}
+
+func EncodeJsonValue(v interface{}) (*JsonValue, error) {
```
andrewpmartinez, line 106: I don't like this. I don't understand why we need this.

controller/model/posture_check_manager.go
plorenz: Yes, I think it happened in one of the acceptance tests with a PATCH

pb/edge_cmd_pb/impl.go
plorenz: It's not used any more and I was going to remove it in a follow up, but I can remove it now

plorenz: removed

andrewpmartinez (approved):

PR #1204, opened 2022-10-11
Add linter and fix issues found by linter

plorenz:


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,41 @@
...
+  pull_request:
+permissions:
+  contents: read
+  # Optional: allow read access to pull request. Use with `only-new-issues` option.
+  # pull-requests: read
+
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
```
andrewpmartinez, line 12: So this will not catch conditional windows/mac only files for lining? Not sure how much we have anymore.

controller/model/posture_check_model_process.go
```diff
@@ -96,8 +96,8 @@ func (p *PostureCheckProcess) FailureValues(_ string, pd *PostureData) PostureCh
 	for _, processData := range pd.Processes {
 		if processData.PostureCheckId == p.PostureCheckId {
 			ret.ActualValue = *processData
+			break
```
andrewpmartinez, line 99: hmmm, this looks like a straight-up bug fix for failure values.

andrewpmartinez (approved):

.github/workflows/golangci-lint.yml
plorenz: I don't think there's much. I think most of what we have is for linux only. I'm ok with missing a few things.

PR #1214, opened 2022-10-21
for #1213 delete api session on authenticator delete/re-enroll

andrewpmartinez:
- an authenticator removed or re-enrolled would continue to allow old API Session to operate
- deleted or re-enrolled authenticators now remove associatd API Sessions


plorenz (commented):
Looks good, but there's a lint error that should be fixed

camotts (approved):

Issue #1226, opened 2022-11-04
Cannot connect to Edge Routers in scenarios using 3rd Party CAs and external claims


andrewpmartinez:
From conversation: https://openziti.discourse.group/t/troubleshoot-error-invalid-client-certificate-for-api-session/824

*Experienced*

Using multiple different 3rd Party CA-issued certificates creates scenarios where either certificate cannot be used for connections to edge routers when using x509 Claims for `externalId` resolution.

*Expected* 

Should not cause issues w/ connections to connections to Edge Routers and Services

*Reproduction*
1) Create a new PKI and add it as a 3rd Party Ca, ensure verification, ensure x509 Claims map to COMMON_NAME
2) Create a service, add attributes for dial and bind
3) Add hosting identity
4) Add dial identity w/ `externalId` set
5) Create two certificates via 3rd Party CA w/  common name set to the desired `externalId`
6) Start host
7) Start two separate dialing identities

PR #1227, opened 2022-11-04
addresses #1226 api session certificate issues w/ 3rd party cas

andrewpmartinez:
- events during a multi-step transaction were immediately emitting events causing out of order and incomplete events to be received, specifically API Sessions w/o fingerprints
- API Session Certificates created implicitly during 3rd Party CA auth did not properly set all properties
- adds tests


plorenz (changes requested):
linter found a few issues, but otherwise looks good

andrewpmartinez:
> linter found a few issues, but otherwise looks good

Fixed.

plorenz (approved):

PR #1228, opened 2022-11-07
updates edge service manager to not allow encryptionRequired to be altered

andrewpmartinez:


specs/management.yml
```diff
@@ -15962,6 +15962,8 @@ definitions:
         items:
           type: string
       encryptionRequired:
+        description: Describes whether connections must support end-to-end encryption
+          on both sides of the connection. Read-only property, set at create.
```
dovholuknf, line 15966: this 'is' the ServiceCreate. feels like this snippet is contradictory here: "Read-only property, set at create"?

specs/source/management/services.yml
```diff
@@ -291,6 +291,7 @@ definitions:
           type: string
       encryptionRequired:
         type: boolean
+        description: Describes whether connections must support end-to-end encryption on both sides of the connection. Read-only property, set at create.
```
dovholuknf, line 294: same comment about it being the create and contradictory to read: "Read-only property, set at create". prolly shoulda commented here

controller/model/edge_service_manager.go
```diff
@@ -65,6 +65,9 @@ func (self *EdgeServiceManager) ApplyCreate(cmd *command.CreateEntityCommand[*Se
 }
 
 func (self *EdgeServiceManager) Update(entity *Service, checker fields.UpdatedFields) error {
+	if checker != nil {
+		checker = checker.RemoveFields("encryptionRequired")
```
dovholuknf, line 69: this is the sorta thing I would log in the controller, but it definitely would be something I'd return back through the API if we can?

dovholuknf (approved):
code is fine and dandy. couple of comments that could/should/might be addressed but - all in all lgtm

camotts (approved):

specs/management.yml
andrewpmartinez: Yeap, find and replace got me. Fixed in a second.

controller/model/edge_service_manager.go
andrewpmartinez: That doesn't make sense. Why would we return anything on this?

dovholuknf: return back meaning: "hey, you asked us to patch this thing, but you sent in a field that we aren't gonna patch". it's confusing to use an API, issue this patch and have it succeed. This PR makes me think that sort of update would silently succeed. sure, someone can just go 'read the docs' but it's more friendly to doc it well, and return a code that says "it mostly worked". like a 206

andrewpmartinez: That will break clients looking for anything else as 206 is not in the spec.

If you are willing to break backward compact, it is better to simply remove it from the spec and cause a code-level break.

PR #1258, opened 2022-11-30
add ebpf mode to edge-router tunnel options

r-caamano:
Created a new mode for edge-router to work with an ebpf based interception FW.  This new mode will make system calls to insert/delete rules into the ebpf interception engine.


tunnel/intercept/ebpf/ebpf_linux.go
```diff
@@ -0,0 +1,520 @@
...
+	interceptAddr.TproxySpec = []string{
+		"-m", "comment", "--comment", service.Name,
+		"-d", ipNet.String(),
+		"-p", interceptAddr.Proto(),
+		"--dport", fmt.Sprintf("%v:%v", interceptAddr.LowPort(), interceptAddr.HighPort()),
+		"-j", "TPROXY",
+		"--tproxy-mark", "0x1/0x1",
+		fmt.Sprintf("--on-ip=%s", port.GetIP().String()),
+		fmt.Sprintf("--on-port=%d", port.GetPort()),
+	}
```
plorenz, line 393: Probably would want to consolidate this logging into a single statement. 
`pfxlog.Logger().WithField("dst_ip", ipNet).WithField("protocol", interceptAddr.Proto()).<rest of fields>.Info("setting up udp intercept)`

tunnel/intercept/ebpf/ebpf_linux.go
```diff
@@ -0,0 +1,520 @@
...
+}
+
+func (self *interceptor) Stop() {
+	self.serviceProxies.IterCb(func(key string, proxy *eBpf) {
+		proxy.Stop(alwaysRemoveAddressTracker{})
+	})
+	self.serviceProxies.Clear()
+}
+
+func (self *interceptor) Intercept(service *entities.Service, resolver dns.Resolver, tracker intercept.AddressTracker) error {
```
plorenz, line 98: Might want to add a check here to make sure that any command line utility you need is present, executable, has the right version, etc, and panic if there's a problem.

r-caamano: Makes sense thanks for the feedback

r-caamano: Hi Paul I added error checking for external map_update binary and associated ebpf maps in "func New()" since that is more consistent with where we check if iptables is available in tproxy mode.

tunnel/intercept/ebpf/ebpf_linux.go
r-caamano: Hi Paul I consolidated logging statements as recommended and also refactored intercept add/delete removing protocol conditional statements since it was redundant code.

PR #1276, opened 2022-12-20
Adding additional info fields to swagger specs

sabedevops:
The additional `info` fields are used by [openapi-generator](https://github.com/OpenAPITools/openapi-generator) client generators for things like module documentation, and packaging scripts.

Example: [setup.py](https://github.com/OpenAPITools/openapi-generator/blob/5300bff6d96adfcfdd0db25b77ecb51532264575/modules/openapi-generator/src/main/resources/python-prior/setup.mustache)

### DIFF
--- a/specs/client.yml
+++ b/specs/client.yml
@@ -7,7 +7,14 @@ schemes:
 swagger: "2.0"
 info:
   title: Ziti Edge Client
-  contact: {}
+  description: OpenZiti Edge Client API
+  contact:
+    name: OpenZiti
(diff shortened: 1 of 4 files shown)


andrewpmartinez (changes requested):
These are generated files and shouldn't (generally) be hand edited. Your changes will be lost.

The files you want are in /specs/source/*.yml.

You can either hand edit both files or edit the files in specs/source/*yml and then assemble them via `README.md` file.

andrewpmartinez (approved):

PR #1283, opened 2023-01-03
Ensure entity counts are on leader only by default. Fix handler remov…

### LINKED ISSUE #1282: Ensure entity count events can be configured to only be emitted on the leader

plorenz:
…al. Fixes #1282


andrewpmartinez:
Got some build errors here.

PR #1291, opened 2023-01-09
support configurable tproxy diverters

scareything:
This change makes it possible to configure a mechanism other than `iptables` to control which destination addresses are diverted to the tunneler's tproxy listener.

The diverter is invoked with the following options:

| activity | options |
|---------|--------|
| router initialization | -V |
| service addition | -I -c `intercept_ip` -m `intercept_prefix` -p `intercept_protocol` -l `intercept_low_port` -h `intercept_high_port` -t `tproxy_listener_port` |
| service removal | -D -c `intercept_ip` -m `intercept_prefix` -p `intercept_protocol` -l `intercept_low_port` -h `intercept_high_port` |

This PR supersedes #1258, as the `map_update` utility can be specified in the tunneler binding's `mode` field.

### DIFF
--- a/tunnel/intercept/tproxy/tproxy_linux.go
+++ b/tunnel/intercept/tproxy/tproxy_linux.go
@@ -36,7 +36,6 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 	"net"
-	"net/netip"
 	"os/exec"
 	"strings"
 	"syscall"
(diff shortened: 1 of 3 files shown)


dovholuknf:
Is there any doc that can be added around this -- or should we add a whole separate issue addressing any related doc?

tunnel/intercept/tproxy/tproxy_linux.go
```diff
@@ -467,35 +493,62 @@ func (self *tProxy) addInterceptAddr(interceptAddr *intercept.InterceptAddress,
...
+		tproxyAddr = self.tcpLn.Addr().(*net.TCPAddr).AddrPort()
+	case "udp":
+		tproxyAddr = self.udpLn.LocalAddr().(*net.UDPAddr).AddrPort()
+	default:
+		return errors.Errorf("unknown protocol %s for service %s", interceptAddr.Proto(), service.Name)
+	}
+
+	if self.interceptor.diverter != "" {
+		cidr := strings.Split(ipNet.String(), "/")
+		cmd := exec.Command(self.interceptor.diverter, "-I",
```
dovholuknf, line 509: cidr[0] and cidr[1] are directly referenced but I don't remember seeing any validation that cidr won't be nil and contains a slash in it. should you add some validation here to prevent a possible panic on user error?

dovholuknf (approved):
lgtm... don't forget the ziti changelog entry :)

tunnel/intercept/tproxy/tproxy_linux.go
scareything: I don't think it could happen, but checking seems wise and easy.

PR #1298, opened 2023-01-24
Use.edge api

andrewpmartinez:
Should be go.mod/sum changes and import statement adjustments.


tests/endpoints
```diff
@@ -0,0 +1 @@
+Endpoints: ['tls:127.0.0.1:6262']
```
dovholuknf, line 1: this change on purpose? i expect it is, but figured i'd ask

dovholuknf (approved):

plorenz (approved):

tests/endpoints
andrewpmartinez: Errant file. Will remove.

andrewpmartinez: It looks like a test is creating it...

PR #1304, opened 2023-02-06
Fix heartbeat deadlock. Fixes #1303

### LINKED ISSUE #1303: Fix deadlock when flushing api session heartbeats

plorenz:
bbolt batch was trying to finish commit, needed all txes to finish heartbeat update in read tx was waiting for concurrent status map so it could update heartbeat status map was locked iterating waiting to submit to batch (now we're back at the beginning)
Was also a view tx within a view tx, which can be problematic


camotts (approved):

dovholuknf (approved):
Lgtm

PR #1357, opened 2023-03-20
reworks broker/sync connect/disconnect logic

andrewpmartinez:
- defensive checks for closed control channels
- defensive checks for duplicate/mismatch control channels

### DIFF
--- a/controller/sync_strats/sync_instant.go
+++ b/controller/sync_strats/sync_instant.go
@@ -157,17 +157,30 @@ func (strategy *InstantStrategy) Stop() {
 }
 
 func (strategy *InstantStrategy) RouterConnected(edgeRouter *model.EdgeRouter, router *network.Router) {
+	log := pfxlog.Logger().WithField("sync_strategy", strategy.Type()).
+		WithField("routerId", router.Id).
+		WithField("routerName", router.Name).
+		WithField("routerFingerprint", *router.Fingerprint)
+
+	//connecting router has closed control channel
+	if router.Control.IsClosed() {
+		log.Errorf("connecting router has closed control channel [id: %s], ignoring", router.Id)
+		return
+	}
+
+	existingRtx := strategy.rtxMap.Get(router.Id)
+
+	//same channel, do nothing
+	if existingRtx != nil && existingRtx.Router.Control == router.Control {
+		log.Errorf("duplicate router connection detected [id: %s], channels are the same, ignoring", router.Id)
+		return
+	}
+
 	rtx := newRouterSender(edgeRouter, router, strategy.RouterTxBufferSize)
 	rtx.SetSyncStatus(env.RouterSyncQueued)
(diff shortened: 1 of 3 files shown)


plorenz (approved):

controller/sync_strats/sync_instant.go
```diff
@@ -157,28 +157,56 @@ func (strategy *InstantStrategy) Stop() {
...
+		log.Infof("edge router [%s] disconnect event, but no rtx found, ignoring", router.Id)
+		return
+	}
+
+	if existingRtx.Router.Control != router.Control && !existingRtx.Router.Control.IsClosed() {
+		log.Infof("edge router [%s] disconnect event, but channels do not match and existing channel is still open, ignoring", router.Id)
+		return
+	}
+
+	log.Infof("edge router [%s] disconnect event, router rtx removed", router.Id)
```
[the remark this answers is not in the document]
andrewpmartinez: It isn't new. The old function immediately outputs an `Infof.` So no more chatty than it is, and it will only happen when ERs are disconnecting.

controller/sync_strats/sync_instant.go
```diff
@@ -157,28 +157,56 @@ func (strategy *InstantStrategy) Stop() {
...
+		WithField("routerId", router.Id).
 		WithField("routerName", router.Name).
-		WithField("routerFingerprint", router.Fingerprint).
-		Infof("edge router [%s] disconnecting", router.Id)
+		WithField("routerFingerprint", *router.Fingerprint)
+
+	existingRtx := strategy.rtxMap.Get(router.Id)
+
+	if existingRtx == nil {
+		log.Infof("edge router [%s] disconnect event, but no rtx found, ignoring", router.Id)
```
dovholuknf, line 199: feels like a metric maybe instead?

controller/sync_strats/sync_instant.go
```diff
@@ -157,28 +157,56 @@ func (strategy *InstantStrategy) Stop() {
...
+		return
+	}
+
+	if existingRtx.Router.Control != router.Control && !existingRtx.Router.Control.IsClosed() {
+		log.Infof("edge router [%s] disconnect event, but channels do not match and existing channel is still open, ignoring", router.Id)
```
dovholuknf, line 204: same comment wrt metric vs log on this. dunno how useful these are other than logging the exact time when it happened

controller/sync_strats/sync_instant.go
andrewpmartinez: It is an edge case that realistically shouldn't happen.

controller/sync_strats/sync_instant.go
andrewpmartinez: It is an edge case that realistically shouldn't happen. I'd only expect it to happen during high-load and flapping ERs. I don't know what kind of metric this would be.

PR #1365, opened 2023-03-23
Cache api sessions for router/tunneler. Fixes #1364

### LINKED ISSUE #1355: Handle duplicate create terminator requests if create terminator fails
If create terminator fails, check if it's because the terminator already exists

### LINKED ISSUE #1359: Sessions creates should be idempotent
If a session is requested and one already exists for that apiSession/service/session-type, return the existing session instead of a new session.

See if we can also make api-session deletes more efficient.

### LINKED ISSUE #1362: Rate limit terminator creates for router/tunneler

### LINKED ISSUE #1364: Cache api-sessions for tunneler router so we don't need to unnecessarily create new sessions
If we have an existing api-session, let's reuse and reduce the load. Will help with flapping routers and upgrades as they will reconnect before the api-session times out. Store api-session in db so we can re-use it no matter if controller or router goes down.

plorenz:
- Cache api sessions for router/tunneler. Fixes #1364

### DIFF
--- a/controller/model/base_manager.go
+++ b/controller/model/base_manager.go
@@ -30,6 +30,8 @@ import (
 	"reflect"
 )
 
+const annotationsBucketName = "annotations"
+
 type EntityManager interface {
 	models.EntityRetriever[models.Entity]
 	command.EntityDeleter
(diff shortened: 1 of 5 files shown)


controller/model/base_manager.go
```diff
@@ -381,6 +383,31 @@ func (self *baseEntityManager) iterateRelatedEntitiesInTx(tx *bbolt.Tx, id, fiel
 	return nil
 }
 
+func (self *baseEntityManager) Annotate(ctx boltz.MutateContext, entityId string, key, value string) error {
+	entityBucket := self.GetStore().GetEntityBucket(ctx.Tx(), []byte(entityId))
+	if entityBucket == nil {
+		return boltz.NewNotFoundError(self.GetStore().GetEntityType(), "id", entityId)
+	}
+	annotationsBucket := entityBucket.GetOrCreatePath(annotationsBucketName)
```
andrewpmartinez, line 391: So there is a reserved bucket named annotate on every entity that stores a single string?

andrewpmartinez (approved):

PR #1368, opened 2023-03-23
Don't stop/start hosting on reconnect, take advantage of idempotent creates instead. Fixes #1336

### LINKED ISSUE #1336: xgress_edge_tunnel shouldn't stop/start host on control channel reconnect
Instead it should take advantage of idempotent creates and send another create message.

plorenz:


router/xgress_edge_tunnel/tunneler.go
```diff
@@ -138,3 +143,56 @@ func (self *tunneler) Close() error {
...
+		self.ReestablishTerminators()
+	}
+}
+
+func (self *tunneler) ReestablishTerminators() {
+	log := pfxlog.Logger()
+	terminators := self.terminators.Items()
+
+	if len(terminators) > 0 {
+		pfxlog.Logger().Infof("reestablishing %v terminators", len(terminators))
```
dovholuknf, line 171: this really info? it seems more debugish to me. as a consumer of this log, it wouldn't mean much?

router/xgress_edge_tunnel/fabric.go
```diff
@@ -363,8 +365,7 @@ func (self *fabricProvider) establishTerminator(terminator *tunnelTerminator) er
 		return err
 	}
 
-	time.Sleep(250 * time.Millisecond)
-	if terminator.created.Load() {
+	if terminator.WaitForCreated(10 * time.Second) {
 		return nil
 	}
```
dovholuknf, line 371: since i can't comment on line 373 (below)... I'll comment here. 

```
return errors.New("check back later")
```
probably isn't super duper useful if to me if i ended up seeing it?

router/xgress_edge_tunnel/fabric.go
```diff
@@ -517,12 +522,31 @@ type tunnelTerminator struct {
...
+	}
+}
+
+func (self *tunnelTerminator) WaitForCreated(timeout time.Duration) bool {
+	if self.created.Load() {
+		return true
+	}
+	select {
+	case <-self.notifyCreated:
+	case <-time.After(timeout):
```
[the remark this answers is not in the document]
plorenz: Probably doesn't need a log because it's just the timeout until this attempt fails. The attempt will be re-retried. So as a 'temporary' failure, I don't think it needs a log.

router/xgress_edge_tunnel/tunneler.go
plorenz: I'm not fussed either way. I'll switch it to a debug

router/xgress_edge_tunnel/fabric.go
plorenz: I'll change this to something like: current attempt timed out, with some context info

plorenz: It's generally not seen, it's mostly a signal to the retry mechanism. It can show up in some cases, though.

router/xgress_edge_tunnel/fabric.go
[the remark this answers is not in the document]
plorenz: Fixed

router/xgress_edge_tunnel/tunneler.go
plorenz: Fixed

router/xgress_edge_tunnel/fabric.go
plorenz: Fixed

PR #1381, opened 2023-03-28
stop specifying body and header params for deleteMfa

### LINKED ISSUE #1380: deleteMfa operation does not actually require an mfaBody
The mgmt spec says I need to send an mfaBody when deleting the mfa on my identity, but the server responds OK if I don't sent the body.

qrkourier:
closes #1380


andrewpmartinez (changes requested):
This is no longer the source of truth for the API specs. See the `edge-api` repos.

Please abandon.

PR #1440, opened 2023-04-21
entity change events

### LINKED ISSUE #629: emit entity change events
There is no mechanism that I am aware of for an integration to observe ziti entities for change (other than polling the controller API.)

As an integrator, I'd like to be able to 'subscribe' to receive change events such that whenever an entity matching my subscription is modified (created, updated, deleted), then my integration is notified.

Notification mechanism could be via web socket, web hook, or simply written to a journal file which can be monitored by a co-located process (ie filebeat.)

Subscriptions filters: by entity type?  include system entities yes/no?  by change type (creates y/n, updates y/n, deletes y/n)?  I'm not sure if there would need to be some abstract 'level' filter to eliminate noise ... as in, level=primary would only event if an entities main properties were changed and would ignore noise in tertiary properties.

Ideally the event would contain some form of identifier of who initiated the change as well as an indication of the changed values (plus the obvious entity type and id.)

### LINKED ISSUE #1295: Ensure DB migrations work properly in a clustered setup (edge)

plorenz:
- Add edge code needed for entity change events. Fixes #629

### DIFF
--- a/controller/model/session_manager_test.go
+++ b/controller/model/session_manager_test.go
@@ -24,7 +24,7 @@ func (ctx *TestContext) testSessionIdempotency(t *testing.T) {
 	service := ctx.requireNewService()
 	service2 := ctx.requireNewService()
 	service.RoleAttributes = []string{eid.New()}
-	ctx.NoError(ctx.managers.EdgeService.Update(service, nil, change.New().SetSource("test")))
+	ctx.NoError(ctx.managers.EdgeService.Update(service, nil, change.New()))
 
 	ctx.requireNewServicePolicy(persistence.PolicyTypeDialName, ss("#all"), ss("#all"))
 	ctx.requireNewServicePolicy(persistence.PolicyTypeBindName, ss("#all"), ss("#all"))
@@ -69,7 +69,7 @@ func (ctx *TestContext) testSessionIdempotency(t *testing.T) {
 	req.Equal(sessSvc1Dial.Id, sessSvc1Dial3.Id)
 	req.Equal(sessSvc1Bind.Id, sessSvc1Bind3.Id)
 
-	req.NoError(ctx.managers.ApiSession.Delete(apiSession.Id, change.New().SetSource("test")))
+	req.NoError(ctx.managers.ApiSession.Delete(apiSession.Id, change.New()))
 	done, err := ctx.GetStores().EventualEventer.Trigger()
 	ctx.NoError(err)
 
--- a/controller/persistence/api_session_store.go
+++ b/controller/persistence/api_session_store.go
@@ -46,14 +46,14 @@ const (
 
 type ApiSession struct {
 	boltz.BaseExtEntity
-	IdentityId      string
-	Token           string
-	IPAddress       string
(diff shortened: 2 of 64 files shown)


controller/model/session_manager_test.go
```diff
@@ -24,7 +24,7 @@ func (ctx *TestContext) testSessionIdempotency(t *testing.T) {
 	service := ctx.requireNewService()
 	service2 := ctx.requireNewService()
 	service.RoleAttributes = []string{eid.New()}
-	ctx.NoError(ctx.managers.EdgeService.Update(service, nil, change.New().SetSource("test")))
+	ctx.NoError(ctx.managers.EdgeService.Update(service, nil, change.New()))
```
andrewpmartinez, line 27: No "test"?

plorenz: I decided providing entity change data for tests, unless entity change events were being tested, was of little value and just extra maintenance overhead. I decided this the third time I changed the API and had to go through and update all the change usages again.

controller/persistence/api_session_store.go
```diff
@@ -46,14 +46,14 @@ const (
...
 	boltz.BaseExtEntity
-	IdentityId      string
-	Token           string
-	IPAddress       string
-	ConfigTypes     []string
-	MfaComplete     bool
-	MfaRequired     bool
-	LastActivityAt  time.Time
-	AuthenticatorId string
+	IdentityId      string    `json:"identity_id"`
```
andrewpmartinez, line 49: So we are adding json tagging to everything? Also, was there reasoning for snasnake ke vs camel case?

controller/persistence/session_store_test.go
```diff
@@ -234,7 +234,7 @@ func (ctx *TestContext) testUpdateSessions(_ *testing.T) {
 	earlier := time.Now()
 	time.Sleep(time.Millisecond * 50)
 
-	mutateCtx := change.New().SetSource("test").NewMutateContext()
+	mutateCtx := change.New().NewMutateContext()
```
andrewpmartinez, line 237: I don't seem to follow from this alone why "test" is sometimes desirable and others not. Can you explain?

andrewpmartinez (approved):
Approved, I guess? I don't like the JSON tags throughout the model, but I'm not sure why we need it there or why not. Also, the tests setting the source as "test" and other times not as "test" seemed odd. I assume it has to do w/ the identity the tests are running as vs lower-level interactions, but I'm not sure.

controller/persistence/session_store_test.go
plorenz: I just gave up half way through and started ripping them back out. I can take out the rest, if you prefer.

plorenz:
> Approved, I guess? I don't like the JSON tags throughout the model, but I'm not sure why we need it there or why not. Also, the tests setting the source as "test" and other times not as "test" seemed odd. I assume it has to do w/ the identity the tests are running as vs lower-level interactions, but I'm not sure.

I removed all the `SetSourceType("test")` for consistency.

PR #1441, opened 2023-04-24
updates golang sdk to v0.20.0

andrewpmartinez:


router/xgress_edge_tunnel/fabric.go
```diff
@@ -48,6 +49,10 @@ import (
 	"time"
 )
```
plorenz, line 51: I feel like we should add this to foundation. Tiny, but generally useful thing

plorenz (approved):
Looks good. All the pointer values make me a bit nervous, but I'm guessing you checked that they're always set.

router/xgress_edge_tunnel/fabric.go
andrewpmartinez: Looks like it is already there in foundation/util/util.go. It is called `Ptr()`

andrewpmartinez:
> Looks good. All the pointer values make me a bit nervous, but I'm guessing you checked that they're always set.

Most of the pointers come from the generated go-swagger rest_mode. Go-swagger uses `nil` for `nullable` and for `required`/`optional` (to defeat zero values like 0, false). The clients validate the payloads well before they hit our code, so if something is required that is missing, an error is generated higher up in the go-swagger client.

PR #1452, opened 2023-05-01
fixes openziti/ziti#1093

andrewpmartinez:
- heart beat system was never being configured with a control channel


plorenz (approved):

plorenz:
lint is failing, but otherwise, I see how it broke and the fix makes sense.

PR #1486, opened 2023-05-12
fixes edge#1225 broken CA de-dupe logic, adds tests

andrewpmartinez:


controller/config/config.go
```diff
@@ -391,11 +391,14 @@ func CalculateCaPems(caPems *bytes.Buffer) *bytes.Buffer {
 			fingerprint := toHex(hash[:])
 			newPem := pem.EncodeToMemory(block)
 			caPemMap[fingerprint] = newPem
```
andrewpmartinez, line 393: This map was calculating fingerprints to de-dupe, it was just never used.

plorenz (approved):

PR #1510, opened 2023-05-31
Fix double terminator remove on channel close. Fixes #1509

### LINKED ISSUE #1509: SDK hosted terminators are being removed twice, causing spurious errors

plorenz:


router/xgress_edge/fabric.go
```diff
@@ -77,6 +77,8 @@ func (self *edgeTerminator) close(notify bool, reason string) {
 			logger.Error("no controller available, unable to remove terminator")
 		} else if err := self.edgeClientConn.removeTerminator(ctrlCh, self); err != nil {
 			logger.WithError(err).Error("failed to remove terminator")
+		} else {
+			logger.Info("Successfully removed terminator on channel close")
```
dovholuknf, line 81: a. this really necessary?
b. if it's necessary, i feel like it should have the id in the message
c. i worry that this will be spammy and unhelpful without additional info

router/xgress_edge/hosted.go
```diff
@@ -41,15 +41,13 @@ func (registry *hostedServiceRegistry) Delete(hostId string) {
...
 	registry.services.Range(func(key, value interface{}) bool {
-		listener := value.(*edgeTerminator)
-		if listener.edgeClientConn == proxy {
-			listener.close(false, "") // don't notify, channel is already closed, we can't send messages
+		terminator := value.(*edgeTerminator)
+		if terminator.edgeClientConn == proxy {
+			terminator.close(false, "") // don't notify, channel is already closed, we can't send messages
 			registry.services.Delete(key)
-			listeners = append(listeners, listener)
 		}
```
dovholuknf, line 50: should this return at this point? will there ever be more than one in the list that matches?

dovholuknf (approved):

router/xgress_edge/fabric.go
plorenz: This is moving the message from the deleted code to here, to maintain the existing logging messages. The logger has the terminator id in it, added at the top of the function.

router/xgress_edge/hosted.go
plorenz: The same sdk connection can have multiple binds on it, so we need to check them all. I had to look it up, though :)

PR #1522, opened 2023-06-13
fixes #1416 allows totp provisioning URL domain to be configured

### LINKED ISSUE #1416: Allow MFA token name to be configured
Right now, when enabling an identity with MFA TOTP, "ziti.dev" is used for the token and relayed to the authenticator apps. It'd be nice to be able to configure that value.

andrewpmartinez:
- sets the default to openziti.io (previously ziti.dev)
- adds `edge.totp` configuration section
- `edge.totp.hostname` now used in MFA TOTP provisioning URL generation

### DIFF
--- a/controller/config/config.go
+++ b/controller/config/config.go
@@ -25,6 +25,7 @@ import (
 	"github.com/openziti/identity"
 	"github.com/pkg/errors"
 	"net"
+	"net/url"
 	"os"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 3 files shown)


plorenz (approved):
Looks good, but there's a related test failing that needs to be updated with the new default URL

andrewpmartinez:
> Looks good, but there's a related test failing that needs to be updated with the new default URL

Oh shoot. I reverted the test files. Will re-add.

PR #1547, opened 2023-07-11
Allow multiple enrollments on identity create. Fixes #1546

### LINKED ISSUE #1546: REST mgmt: create identity - no ottca enrollment created when both ottca and ott are true
This request results in an identity with one `ott` enrollment. The requested `ottca` enrollment is not created.

`POST /edge/management/v1/identities`

```json
{
  "name": "kentest25",
  "type": "Device",
  "isAdmin": false,
  "appData": {},
  "authPolicyId": "default",
  "defaultHostingCost": 0,
  "defaultHostingPrecedence": "required",
  "enrollment": {
    "ott": true,
    "ottca": "5L5JFo5dxKUFbiNXohMPll",
    "updb": ""
  },
  "externalId": "",
  "roleAttributes": [
    "testapi-clients"
  ],
  "serviceHostingCosts": {},
  "serviceHostingPrecedences": {},
  "tags": {}
}
```

If the same request is sent with `ott: false` the expected `ottca` enrollment is created.

plorenz:


qrkourier (approved):
So, from the mgmt API's perspective, a request might ask for zero or more enrollments including:

* `updb`
* `ott`
* `ottca`

If no enrollments are requested then the identity can only be used with external ID claims unless one of the above enrollments is added later.

andrewpmartinez (approved):
I was considering removing this functionality altogether and letting create identity and create enrollment be separate concerns.

If this functionality is to be maintained, a test would be nice.

plorenz:
> 

I think removing it would be cleanest. We had the same thing with services and terminators, but I think I stripped that out a while ago. 
I think the CLI and UIs should probably still make it one action, just with two REST calls under the hood.

plorenz:
Closing in favor of removing nested enrollments in identity create, and requiring explicit enrollment creation.

PR #1557, opened 2023-07-20
adds support OIDC authentication, api access, and ER connections

andrewpmartinez:


controller/model/revocation_manager.go
```diff
@@ -0,0 +1,103 @@
...
+
+	network.RegisterManagerDecoder[*Revocation](env.GetHostController().GetNetwork().Managers, manager)
+
+	return manager
+}
+
+type RevocationManager struct {
+	baseEntityManager[*Revocation, *persistence.Revocation]
+}
+
```
plorenz, line 45: You can avoid this if you use `RegisterCreateDecoder` and `RegisterDeleteDecoder` instead of `RegisterManagerDecoder`.

pb/edge_ctrl_pb/edge_ctrl.proto
```diff
@@ -109,6 +111,46 @@ message Service {
   bool encryptionRequired = 3;
 }
```
plorenz, line 113: I see some of these types used, but not all. Are some of them for future work or intermediary stuff that could be tidied?

plorenz (approved):
Couple of comments, but overall looks good to me

pb/edge_ctrl_pb/edge_ctrl.proto
andrewpmartinez: They are for follow-up functionality on a different branch.

PR #1598, opened 2023-08-30
Put api session into tunnel state after looking up from cache. Fixes #1597

### LINKED ISSUE #1597: ER/T cached API sessions aren't remembered when pulled from cache
ER/T cached API sessions aren't remembered when pulled from cache. If we are able to re-use an api session for an ER/T context, we're not storing it back into the context, meaning we look it up every time we are dialing or hosting a service. This is extra work that doesn't need to be done and results in unnecessary notifications back to the router that the api session has changed.

plorenz:


dovholuknf (approved):
lgtm!

PR #1614, opened 2023-09-15
filter intercepted clients by source ip

scareything:


plorenz (approved):

plorenz:
The edge project is being archived. This branch will need to be moved to the ziti project if we want to keep it.

PR #1615, opened 2023-09-15
fixes openziti/ziti#1310 enrollment extension waits for controller

andrewpmartinez:
For router certificate extension, routers now wait for a control channel for 10s then check again every one minute till it is available or the router is terminated or self terminates due to a lack of a control channel.


plorenz (approved):

qrkourier (approved):
will test with `openziti/ziti-controller:release-next` container after merge

